PR #11, opened 2022-08-05
Fix syntax error in logstream put handler

trueleo:
This PR fixes a syntax error on line 181 of handlers/logstream.rs


nitisht (changes requested):
@trueleo can you fix rust fmt issue

nitisht (approved):

PR #14, opened 2022-08-06
Move authentication middleware to new scope

trueleo:
### Description

BasicAuth middleware was wrapped around entire application causing people to authenticate twice. 
This PR fixes this my moving all endpoints under base path to a new scope then wrapping it in the BasicAuth middleware  
Now the static files served under root path do not require authentication.


server/src/main.rs
```diff
@@ -188,26 +193,27 @@ macro_rules! create_app {
...
+        "/logstream".to_string()
+    } else {
+        format!("/logstream/{}", stream_name)
     }
-    format!("{}/logstream/{}", base_path(), stream_name)
 }
 
 fn readiness_path() -> String {
-    format!("{}/readiness", base_path())
+    "/readiness".to_string()
```
nitisht, line 208: How will it get the base path?

server/src/main.rs
```diff
@@ -188,26 +193,27 @@ macro_rules! create_app {
...
 }
 
 fn liveness_path() -> String {
-    format!("{}/liveness", base_path())
+    "/liveness".to_string()
```
nitisht, line 212: same as above

server/src/main.rs
```diff
@@ -188,26 +193,27 @@ macro_rules! create_app {
...
 }
 
 fn query_path() -> String {
-    format!("{}/query", base_path())
+    "/query".to_string()
```
nitisht, line 216: same as above

nitisht (changes requested):

server/src/main.rs
```diff
@@ -135,40 +134,46 @@ async fn run_http() -> anyhow::Result<()> {
...
-        )
-        // GET "/liveness" ==> Livenss check as per https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-command
-        .service(web::resource(liveness_path()).route(web::get().to(handlers::liveness)))
-        // GET "/readiness" ==> Readiness check as per https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-readiness-probes
-        .service(web::resource(readiness_path()).route(web::get().to(handlers::readiness)))
-        // GET "/" ==> Serve the static frontend directory
-        .service(ResourceFiles::new("/", generated));
+    cfg.service(
+        // Base path "{url}/api/v1"
+        web::scope(&base_path())
```
trueleo, line 139: @nitisht base_path is used in this scope thus every path grouped underneath it will share same base url i.e /api/v1. I changed the related functions to not return full path.

nitisht: ah okay..we need to test all the APIs then, just to make sure nothing is broken

trueleo: Yep

nitisht (approved):

ProgrammerShri:
Looking forward to it...
@trueleo 

Good Job

PR #19, opened 2022-08-07
Fix auth issue

### LINKED ISSUE #17: [UI] do not store username and password in local store
We don't need to store the username and password in local store. Let's remove that code.

ameen4455:
Fixes #17

This PR has:


ui/src/page/Login/index.js
```diff
@@ -31,12 +31,9 @@ const Login = () => {
...
       localStorage.setItem("auth", credentials);
       localStorage.setItem("username", username);
-      localStorage.setItem("password", password);
       try {
-        const res = await getLogStream();
-        if (res.statusText === "OK") {
-          navigate("/index.html");
-        }
+        await getLogStream();
+        navigate("/index.html");
```
nitisht, line 36: But this would mean no check at all right? If the `getLogStream()` function failed wouldn't this still navigate to `index.html`

ameen4455: If the request fails, it would be caught by the try catch block

nitisht: so anything other than 200 https status would be a failure here?

ameen4455: 4XX would be considered failure

nitisht: okay, then we're good.

nitisht (approved):

PR #24, opened 2022-08-09
Fix build process for unified binary

### LINKED ISSUE #18: [Server] New approach to separate ui code from server and still build a unified binary
Current approach is difficult to maintain and requires rebuild both ui and rust components for every ui change. We need to find a new approach that helps us create stable builds with much less developer effort.

trueleo:
Fixes #18.

### Description

This PR allows for UI to be decoupled from the parseable repo and yet be able to merge server and frontend into unified binary. In production environment build script will look for `assets-url` and `assets-sha1` in package metadata specified in  Cargo.toml, this asset will then be downloaded and linked as resource when binary is being built.

An environment variable called `LOCAL_ASSETS_PATH` is provided for development purposes. if this is set to a valid path ( i.e build folder of the frontend ) before running cargo then it will include UI artefacts directly from build folder specified.

### DIFF
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -54,7 +54,6 @@ static-files = "0.2.1"
 cargo_toml = "0.11.5"
 ureq = "2.5.0"
 sha1_smol = {version = "1.0.0", features=["std"]}
-hex = "0.4.3"
 zip = { git = "https://github.com/zip-rs/zip" }
 
 [package.metadata.parseable_ui]
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -57,8 +57,8 @@ sha1_smol = {version = "1.0.0", features=["std"]}
 zip = { git = "https://github.com/zip-rs/zip" }
 
 [package.metadata.parseable_ui]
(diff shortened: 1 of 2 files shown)


server/Cargo.toml
```diff
@@ -50,4 +50,17 @@ static-files = "0.2.1"
...
+cargo_toml = "0.11.5"
+ureq = "2.5.0"
+sha1_smol = {version = "1.0.0", features=["std"]}
+hex = "0.4.3"
+zip = { git = "https://github.com/zip-rs/zip" }
+
+[package.metadata.parseable_ui]
+assets-url = "http://localhost:5555/ui/build.zip"
+assets-sha1 = "541f1887087ff80149b1d19bce802b4e359cb2e3"
+# specifies where to find local assests
```
nitisht, line 63: Published a build.zip on the frontend repo. 

https://github.com/parseablehq/frontend/releases/tag/v0.0.1

Sha is `d9243bde7a98bdd8a19ed47cca2f12f9a5ce40ba`. Can you please change the links here to the correct links and try building once?

trueleo: Yeah, I'll update it right away

PR #28, opened 2022-08-12
feat(alerts): foundational implementation

### LINKED ISSUE #4: [Server] Implement alerting
We need to add alerting optional per log stream. Config would look like this:

```json
{
  "alerts": [
    {
      "name": "server-fail-alert",
      "message": "server reported error status",
      "rule": {
        "field": "http_status",
        "contains": "500", // equals, greaterThan, lessThan
        "repeats": "5", // repeats this many times
        "within": "10m" // for this duration seconds (s), minutes (m), hours (h), days (d)
      },
      "target": [
        {
          "name": "slack-target",
          "server_url": "http://mailgun.com",
          "api_key": "xxxx"
        }
      ]
    }
  ]
}
```

This config would be added to the Parseable server, per log stream via API

de-sh:
Fixes #4

### Description
Initial implementation of alerting as a feature

<hr>

This PR has:


server/src/alerts.rs
```diff
@@ -0,0 +1,113 @@
+use serde::{Deserialize, Serialize};
```
nitisht, line 1: add license header here

server/src/alerts.rs
```diff
@@ -0,0 +1,113 @@
...
+    }
+}
+
+#[derive(Default, Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Rule {
+    pub field: String,
+    /// Field that determines what comparison operator is to be used
+    #[serde(default)]
+    pub comparator: Comparator,
```
nitisht, line 41: Instead of comparator, lets call this `Operator`

PR #31, opened 2022-08-14
Change stream_name type

trueleo:
### Description
This PR changes stream_name argument type for `metadata::STREAM_INFO::schema` and `metadata::STREAM_INFO::delete_stream`
 
Both of these methods use stream_name argument only for lookup and not for storing of any kind so it is alright for them to be of borrow type. This allows for removal of some unnecessary cloning.
<hr>


nitisht (approved):
Looking good!

PR #32, opened 2022-08-14
Add rstest for testing 

trueleo:
### Description

This PR introduces [rstest](https://lib.rs/crates/rstest) as main unit testing framework. It's added as a development dependency to the project. [maplit](https://docs.rs/maplit/latest/maplit/) is added for its hashmap macro.
 
#### Core advantages of using rstest
- can save on some boilerplate code
- easier to add new test cases
- #[once] Fixture to make an object initialise once and be shared between all tests

#### Disadvatages of using rstest
- Complex struct initialisation might not be possible in a test case. This can still be a no-issue because we can always write normal rust tests for anything that can't be done using rstest.

#### What is changed 
- tests added to metadata.rs
- tests converted in query.rs
- tests converted in utils.rs
 
<hr>

This PR has:

### DIFF
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -56,6 +56,10 @@ ureq = "2.5.0"
 sha1_smol = { version = "1.0.0", features=["std"] }
 zip = { git = "https://github.com/zip-rs/zip" }
 
+[dev-dependencies]
+maplit = "1.0.2"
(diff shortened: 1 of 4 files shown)


nitisht:
@de-sh any comments on the `rstest` use?

PR #45, opened 2022-08-18
Introduce a new error type for ObjectStorage api

### LINKED ISSUE #42: Panic when bucket doesn't exists
Server logs panic error when a bucket as the Parseable storage is not provided.

```
thread 'actix-rt|system:0|arbiter:2' panicked at 'called `Result::unwrap()` on an `Err` value: Storage(NoSuchBucket(NoSuchBucket { message: Some("The specified bucket does not exist") }))', server/src/handlers/logstream.rs:79:60
```

We need to handle the properly and log a better error - and not start the server if bucket doesn't exist.

trueleo:
Fixes #42.

### Description

Create a separate error type for all the methods defined by `ObjectStorage`. This `ObjectStorageError` should contain those failure variants over which we want to handle error in future. 

`AwsSdkError` and `SdkError` in AWS S3 crate covers all the variants for failure when calling S3 APIs but parseable needs to have its own Error type that can be used for error handling in the application itself.

Added `ObjectStorageError` type which contains some variants through which we can tell what went wrong when using any of trait methods on S3 object. Though the error variants need to expand in future so that more complex cases could be handled correctly

<hr>

This PR has:

### DIFF
--- a/server/src/metadata.rs
+++ b/server/src/metadata.rs
@@ -139,14 +139,12 @@ impl STREAM_INFO {
                 .and_then(parse_string)
                 .map_err(|_| Error::SchemaNotInStore(stream.name.to_owned()));
 
-            let metadata = match (alert_config, schema) {
-                (Ok(alert_config), Ok(schema)) => LogStreamMetadata {
-                    schema,
-                    alert_config,
-                    ..Default::default()
-                },
-                _ => continue,
+            let metadata = LogStreamMetadata {
+                schema: schema.unwrap_or_default(),
+                alert_config: alert_config.unwrap_or_default(),
+                ..Default::default()
             };
+
             let mut map = self.write().unwrap();
             map.insert(stream.name.clone(), metadata);
         }
--- a/server/src/metadata.rs
+++ b/server/src/metadata.rs
@@ -128,12 +128,14 @@ impl STREAM_INFO {
             let alert_config = storage
                 .get_alert(&stream.name)
                 .await
+                .map_err(|e| e.into())
                 .and_then(parse_string)
                 .map_err(|_| Error::AlertNotInStore(stream.name.to_owned()));
 
             let schema = storage
                 .get_schema(&stream.name)
                 .await
+                .map_err(|e| e.into())
                 .and_then(parse_string)
                 .map_err(|_| Error::SchemaNotInStore(stream.name.to_owned()));
 
--- a/server/src/metadata.rs
+++ b/server/src/metadata.rs
@@ -18,7 +18,6 @@
 
 use bytes::Bytes;
 use lazy_static::lazy_static;
-use log::warn;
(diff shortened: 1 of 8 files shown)


server/src/option.rs
```diff
@@ -72,6 +73,22 @@ impl Config {
         }
     }
 
+    pub async fn validate_storage(&self, storage: &impl ObjectStorage) {
+        match storage.check().await {
+            Ok(_) => (),
+            Err(ObjectStorageError::NoSuchBucket(name)) => panic!(
+                "Could not start because the bucket named {bucket} doesn't exist. Please make sure bucket with the name {bucket} exists on {url} and then start parseable again",
+                bucket = name,
+                url = self.storage.endpoint_url()
```
trueleo, line 82: This is one of the usage for the error kind which fixes #42. Panic when bucket does not exist or we fail to connect to storage endpoint

Are these error message good ?

nitisht: checking

server/src/option.rs
```diff
@@ -72,6 +73,22 @@ impl Config {
...
+            ),
+            Err(ObjectStorageError::ConnectionError(inner)) => panic!(
+                "Failed to connect to the Object Storage Service\nCaused by: {}",
```
nitisht, line 85: Add the {url} here as well.

server/src/option.rs
```diff
@@ -72,6 +73,22 @@ impl Config {
         }
     }
 
+    pub async fn validate_storage(&self, storage: &impl ObjectStorage) {
+        match storage.check().await {
+            Ok(_) => (),
+            Err(ObjectStorageError::NoSuchBucket(name)) => panic!(
+                "Could not start because the bucket named {bucket} doesn't exist. Please make sure bucket with the name {bucket} exists on {url} and then start parseable again",
+                bucket = name,
```
nitisht, line 81: {bucket} is repeated. Make it like

```
"Could not start because the bucket doesn't exist. Please ensure bucket {bucket} exists on {url}. Then start Parseable."
```

server/src/metadata.rs
```diff
@@ -139,14 +139,12 @@ impl STREAM_INFO {
...
-            let metadata = match (alert_config, schema) {
-                (Ok(alert_config), Ok(schema)) => LogStreamMetadata {
-                    schema,
-                    alert_config,
-                    ..Default::default()
-                },
-                _ => continue,
+            let metadata = LogStreamMetadata {
+                schema: schema.unwrap_or_default(),
+                alert_config: alert_config.unwrap_or_default(),
```
trueleo, line 144: Instead of returning or not adding metadata to stream we add empty string for now. This allows first event to not fail in case parseable was restarted just after adding a stream

nitisht (approved):

PR #49, opened 2022-08-19
Manage sync threads failures

### LINKED ISSUE #43: Mechanism to ensure local data sync thread is always running
The local data sync thread is responsible for moving the data to a local directory and then move it to object storage as needed. If this thread crashes for any reason, we should be able detect that, try restarting the thread. 

If the thread can't be respawned due to whatever reasons, the server should at the end of granularity period should stop taking post events (because we can't ensure data integrity).

trueleo:
Fixes #43

### Description

This PR makes bunch of changes to ways how local sync and s3 sync threads are managed inside parseable. The goal is to make parseable main thread be aware of unwanted failure inside either of these dedicated sync thread. Additionally we want to have control over life cycle of these threads and have them stop on command, This can be later explored in future if we want to have graceful shutdown an also to guarantee more data consistency in case of failure.

#### This solution works something like this :-
Sync threads are spawned with pair of one shot channels (named with suffix inbox and outbox) for communication with main thread.

- inbox is sender type which can be used to stop the respective thread by breaking its scheduler loop.
- outbox is receiver type which is polled inside the main thread's looped select. So that whenever we receive a message from thread we stop.     

scheduler is ran inside loop which runs pending tasks and polls the channel for any message from main thread ( if any then stop the thread )

All of this is wrapped by a catch unwind. So in case of an unwinding because of panic there is a chance to let main thread know and handle accordingly. This is added as a safeguard for now but later we need to verify that local_sync and s3_sync don't have anything that can panic.

<hr>

This PR has:

### DIFF
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -22,13 +22,17 @@ use actix_web::{middleware, web, App, HttpServer};
 use actix_web_httpauth::extractors::basic::BasicAuth;
 use actix_web_httpauth::middleware::HttpAuthentication;
 use actix_web_static_files::ResourceFiles;
-use clokwerk::{AsyncScheduler, TimeUnits};
+use clokwerk::{AsyncScheduler, Scheduler, TimeUnits};
 use log::warn;
 use openssl::ssl::{SslAcceptor, SslFiletype, SslMethod};
 
 include!(concat!(env!("OUT_DIR"), "/generated.rs"));
 
-use std::thread;
+use std::panic::{catch_unwind, AssertUnwindSafe};
+use std::thread::{self, JoinHandle};
+use std::time::Duration;
+use tokio::sync::oneshot;
+use tokio::sync::oneshot::error::TryRecvError;
 
 mod banner;
 mod error;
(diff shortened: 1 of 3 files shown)


server/src/main.rs
```diff
@@ -63,33 +67,110 @@ async fn main() -> anyhow::Result<()> {
...
+                s3sync_inbox.send(()).unwrap_or(());
+                localsync_inbox.send(()).unwrap_or(());
+                localsync_handler.join().unwrap_or(());
+                s3sync_handler.join().unwrap_or(());
+                return e
+            },
+            _ = &mut localsync_outbox => {
+                // crash the server if localsync fails for any reason
+                // panic!("Local Sync thread died. Server will fail now!")
+                return Err(anyhow::Error::msg("Local Sync failed"))
```
nitisht, line 89: Let's improve the error message here. User might not know what is local sync. Let's say

```
Failed to sync local data to disc. This can happen due to critical error in disc or environment. Please restart the Parseable server. 

Join us on Parseable Slack if the issue persists after restart : https://launchpass.com/parseable
```

server/src/main.rs
```diff
@@ -63,33 +67,110 @@ async fn main() -> anyhow::Result<()> {
...
+            scheduler
+                .every((storage::LOCAL_SYNC_INTERVAL as u32).seconds())
+                .run(move || {
+                    if let Err(e) = S3::new().local_sync() {
+                        warn!("failed to sync local data. {:?}", e);
+                    }
+                });
+
+            loop {
+                thread::sleep(Duration::from_millis(50));
```
trueleo, line 155: Is it okay to thread sleep here for small amount of time ?

nitisht: Why is sleep needed here?

PR #50, opened 2022-08-19
Fix query validation to ensure we use actual query value

### LINKED ISSUE #47: It is not possible to query fields with capitals names
Query is case sensitive, so we need to avoid modifying query string while processing the query.

nitisht:
This PR fixes the situation where original query sent
by the user was converted to lower case. We now lower case
the query for validation only. Then pass on the
actual query to datafusion.

Fixes #47


trueleo:
Seems good

trueleo (approved):

PR #55, opened 2022-08-24
Refactor StorageSync

### LINKED ISSUE #54: Verify valid streams on startup and make sure stream is properly deleted
When a stream is deleted it is removed from the object store and metadata. In case of it being an active stream there might be parquet files for it in local directory waiting to be synced or might already be synced just after deletion. This behaviour causes it to be loaded as a stream again because a root level prefix is created on storage by s3 sync. 

Possible fixes
- Check for .schema on each stream when loading for metadata
- Clean up code that cleans invalid streams if any

trueleo:
### Description
This PR refactors code around `StorageSync`, fixes instances where `Pathbuf` is more appropriate type to for path handling and changes filename such that it adds hostname for identification. 

StorageSync is only ever required for local sync cycle. On s3 sync currently it checks for top level folder inside local data directory to retrieve stream names and respective paths, this is not ideal so this is changed so that we only go through streams that are returned through 'list_streams' in s3 storage. This partiallty fixes #54 but there should be more checks in place when loading streams.

#### Changes
 - Refactors StorageSync
 - Change from DirName to StorageDir for path related stuff
 - Use StorageDir in s3 sync
 - Only check for files on depth 1 when walking tmp directory 
 - use hostname for file instead of random string
 
<hr>

This PR has:

### DIFF
--- a/server/src/metadata.rs
+++ b/server/src/metadata.rs
@@ -170,6 +170,10 @@ impl STREAM_INFO {
         Ok(())
     }
 
+    pub fn list_streams(&self) -> Vec<String> {
+        self.read().unwrap().keys().map(String::clone).collect()
+    }
+
     pub fn update_stats(
         &self,
         stream_name: &str,
--- a/server/src/storage.rs
+++ b/server/src/storage.rs
@@ -149,7 +149,6 @@ pub struct LogStream {
 
 #[derive(Debug)]
 struct StorageDir {
-    stream_name: String,
     pub data_path: PathBuf,
     pub temp_dir: PathBuf,
 }
@@ -160,20 +159,9 @@ impl StorageDir {
         let temp_dir = data_path.join("tmp");
 
         Self {
-            stream_name,
             data_path,
             temp_dir,
         }
-    }
-
-    #[allow(dead_code)]
-    fn temp_dir_path(&self) -> PathBuf {
-        self.temp_dir.clone()
-    }
-
-    #[allow(dead_code)]
-    fn local_stream_data_path(&self) -> PathBuf {
-        CONFIG.parseable.local_stream_data_path(&self.stream_name)
     }
 
     fn create_temp_dir(&self) -> io::Result<()> {
--- a/server/src/storage.rs
+++ b/server/src/storage.rs
@@ -17,7 +17,7 @@
  */
 
 use crate::alerts::Alerts;
-use crate::metadata::Stats;
+use crate::metadata::{Stats, STREAM_INFO};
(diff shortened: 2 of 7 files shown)


server/src/storage.rs
```diff
@@ -114,35 +111,30 @@ pub trait ObjectStorage: Sync + 'static {
...
-                    if let Err(e) = dir.delete_parquet_file(file_local.clone()) {
-                        log::error!(
-                            "Error deleting parquet file in path {} due to error [{}]",
-                            file_local,
-                            e
-                        );
-                    }
+                let filename = file.file_name().unwrap().to_str().unwrap();
+                let file_suffix = str::replacen(filename, ".", "/", 3);
+                let s3_path = format!("{}/{}", stream, file_suffix);
```
nitisht, line 129: lets use the `join` approach everywhere

server/src/storage.rs
```diff
@@ -114,35 +111,30 @@ pub trait ObjectStorage: Sync + 'static {
...
-                    let _put_parquet_file = self.upload_file(&f_new_path, &file_local).await?;
-                    if let Err(e) = dir.delete_parquet_file(file_local.clone()) {
-                        log::error!(
-                            "Error deleting parquet file in path {} due to error [{}]",
-                            file_local,
-                            e
-                        );
-                    }
+                let filename = file.file_name().unwrap().to_str().unwrap();
+                let file_suffix = str::replacen(filename, ".", "/", 3);
```
nitisht, line 128: lets avoid adding new unwraps

nitisht (commented):
some comments

server/src/metadata.rs
```diff
@@ -168,6 +168,10 @@ impl STREAM_INFO {
         Ok(())
     }
 
+    pub fn list_streams(&self) -> Vec<String> {
+        self.read().unwrap().keys().map(String::clone).collect()
+    }
```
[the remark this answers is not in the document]
trueleo: unwrapping here is fine because same thread does not hold any lock. Some measures can be considered to handle poisoning across all instances where read is needed.

server/src/storage.rs
trueleo: file is a valid utf-8 filename generated when local sync happens. Only way this could fail is if someone created an invalid file in the tmp directory.

nitisht: Someone could try this, we can't underestimate our users :)

server/src/storage.rs
nitisht: right

PR #56, opened 2022-08-24
Alert timing

de-sh:
Fixes #XXXX.

### Description

Adds code to manage time periods within which the alert rule is valid.

Possible loopholes: The current behavior only considers first event only for setting `period_end`, so lets say the range is 5s and there are events that come in at 0s, 2s, 3s, 4s, 5s, 6s, 7s, at 6s, reset will cause loss of information, and we will have to restart from 6s until 11s or so on. Could this be bad design?

<hr>

This PR has:


nitisht:
Closing this for now. We'll revisit when cleaning up the alert feature.

PR #57, opened 2022-08-24
Update Parseable Helm Chart

AdheipSingh:
### Description
- Commenting the S3 env configs, as they get deployed as empty strings, and server takes those into consideration
- Default should be commented off.
- Added RUST_LOG to get logs for server
- update Dockerfile rust version
- Added security context

<hr>

This PR has:

### DIFF
--- a/helm/parseable/Chart.yaml
+++ b/helm/parseable/Chart.yaml
@@ -3,4 +3,4 @@ name: parseable
 description: Helm chart for Parseable Server
 type: application
 version: 0.0.1
-AppVersion: ""
+AppVersion: "sha-b6d9366"
--- a/helm/parseable/values.yaml
+++ b/helm/parseable/values.yaml
@@ -9,11 +9,13 @@ parseable:
     P_TLS_CERT_PATH: ""
     P_LOCAL_STORAGE: "data"
     P_STORAGE_UPLOAD_INTERVAL: "60"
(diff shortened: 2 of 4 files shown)


helm/parseable/values.yaml
```diff
@@ -9,11 +9,13 @@ parseable:
     P_TLS_CERT_PATH: ""
     P_LOCAL_STORAGE: "data"
     P_STORAGE_UPLOAD_INTERVAL: "60"
-    P_S3_URL: ""
-    P_S3_ACCESS_KEY: ""
-    P_S3_SECRET_KEY: ""
-    P_S3_REGION: ""
-    P_S3_BUCKET: ""
+    RUST_LOG: "parseable=trace"
```
nitisht, line 12: IMO we shouldn't enable this by default..

helm/parseable/Chart.yaml
```diff
@@ -3,4 +3,4 @@ name: parseable
 description: Helm chart for Parseable Server
 type: application
 version: 0.0.1
-AppVersion: ""
+AppVersion: "sha-b6d9366"
```
nitisht, line 6: This should be a proper release version, we can update once we make the `0.0.1` release?

AdheipSingh: yes once release is cut, we have to bump versions

PR #72, opened 2022-08-28
Fix logstream listing and deletion

### LINKED ISSUE #54: Verify valid streams on startup and make sure stream is properly deleted

trueleo:
Fixes #54 .

- #### Delete local data when stream is deleted.

- #### Use `/.schema` as prefix when listing streams.
    As s3 sync is an independent thread there is no guarantee of ordering between when stream list is read and when it is deleted. However when a stream is deleted then it's .schema is gone. So by checking for .schema valid streams can be listed.

<hr>

This PR has:

### DIFF
--- a/server/src/handlers/logstream.rs
+++ b/server/src/handlers/logstream.rs
@@ -16,13 +16,15 @@
  *
  */
 
+use std::fs;
+
 use actix_web::http::StatusCode;
 use actix_web::{web, HttpRequest, HttpResponse, Responder};
 
 use crate::alerts::Alerts;
 use crate::response;
 use crate::s3::S3;
-use crate::storage::ObjectStorage;
+use crate::storage::{ObjectStorage, StorageDir};
 use crate::{metadata, validator};
 
 pub async fn delete(req: HttpRequest) -> HttpResponse {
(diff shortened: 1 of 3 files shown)


server/src/handlers/logstream.rs
```diff
@@ -46,21 +48,30 @@ pub async fn delete(req: HttpRequest) -> HttpResponse {
         .to_http();
     }
 
-    if let Err(e) = s3.delete_stream(&stream_name).await {
+    if let Err(e) = metadata::STREAM_INFO.delete_stream(&stream_name) {
```
nitisht, line 51: order of deletion matters here. First delete from s3 because that has highest probability to fail. Only on successfully deleting from s3, attempt local directory and then metadata

If s3 deletion fails, fail the api
if s3 delete worked and local dir delete failed, respond with 200 in api response and log on console about failed delete in local data dir.
if s3 and local dir delete worked and metadata delete failed, same as second case, api is successful and error log on console about exact failure

PR #73, opened 2022-08-28
Move leftover data.parquet to tmp with valid prefix on server startup

### LINKED ISSUE #53: Contents in `data.parquet` need to be preserved properly in case of restart
When the server restarts, the left over `data.parquet` file needs to be properly tagged with relevant time stamp (when it was last modified) and moved to the relevant directory.

Then server should start with a fresh `data.parquet` file.

trueleo:
Fixes #53 .

### Description

When server is restarted it should try to sync data on the local directory that is not yet synced to s3. Otherwise the data will get lost when new event arrives.

Make use of last modified system time to generate a prefix and move `data.parquet` to tmp directory, It will be synced to s3 on next `s3_sync` run. 
 
<hr>

This PR has:

### DIFF
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -22,8 +22,9 @@ use actix_web::{middleware, web, App, HttpServer};
 use actix_web_httpauth::extractors::basic::BasicAuth;
 use actix_web_httpauth::middleware::HttpAuthentication;
 use actix_web_static_files::ResourceFiles;
-use chrono::{DateTime, Timelike, Utc};
+use chrono::{DateTime, NaiveDateTime, Timelike, Utc};
 use clokwerk::{AsyncScheduler, Scheduler, TimeUnits};
+use filetime::FileTime;
 use log::warn;
 use openssl::ssl::{SslAcceptor, SslFiletype, SslMethod};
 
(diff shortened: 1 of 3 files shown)


server/src/main.rs
```diff
@@ -98,6 +105,65 @@ async fn main() -> anyhow::Result<()> {
...
+        return;
+    }
+
+    for stream in metadata::STREAM_INFO.list_streams() {
+        let dir = StorageDir::new(&stream);
+        // if data.parquet file is not present then skip this stream
+        if !dir.parquet_path_exists() {
+            continue;
+        }
+        if let Err(e) = dir.create_temp_dir() {
```
nitisht, line 119: what is the behaviour here if the `tmp` directory exists already? I mean we need to ensure we don't overwrite, because it is highly probably that the directory exists and has some data that is still not sent to S3.

trueleo: create_dir_all  recursively builds all the missing component of path. If there is already an existing tmp, it will not overwrite anything inside

nitisht (commented):
This is not working on macOS and mostly windows as well. For platforms where we can't fetch the last modified time, we still need to discard data.parquet if that exists in the `./data` directory. Otherwise we'd be adding new data to stale data from when the server crashed

trueleo:
@nitisht Test for this again after #77 is merged. Issue might be that the logstream was not getting loaded so there was no try on startup sync

nitisht:
seems to be working now.

PR #79, opened 2022-08-29
Collect label and metadata from header

### LINKED ISSUE #75: Console needs to decouple from a specific schema
Console needs to ensure any schema that user has sent to a stream is properly visible and understandable in the dashboard. Those changes are going on in the console repo. But there two specific things backend needs to ensure:

All the log data that is sent as body in `POST` request gets stored as is in the parquet file. But server should have two special columns in each stream (also stored in the parquet file).

1. `p_metadata`
2. `p_tags`

Both these fields are sent as HTTP headers during the post request. So server would parse any field starting with `x-p-meta:m1=v1,m2=v2` to `p_metadata` and `x-p-tags:t1=v1,t2=v2` to `p_tags`. 

Also both these special fields should be returned in the get schema api call for a stream.

trueleo:
Fixes #75 .

### Description

This PR allows for metadata and tags to be fetched from headers instead of body of the event. This gives some flexibility in schema and multiple sources with different schema for available metadata can send logs to the same stream.  

Client needs to use prefix `x-p-meta-` and `x-p-tags-` in headers to send key value pairs in a server post event. These headers are parsed and validated before being merged into body with `p_metadata` and `p_tags` field respectively

### Changes 
- introduced header_parsing::collect_labelled which parses headers key value pairs and gives a string formatted as "k1=v1,k2=v2"
- New Error type ParseHeaderError which implements ResponseError.
- Use above types in post event handler

<hr>

This PR has:

### DIFF
--- a/server/src/handlers/event.rs
+++ b/server/src/handlers/event.rs
@@ -16,8 +16,10 @@
  *
  */
 
+use std::collections::HashMap;
+
 use actix_web::http::StatusCode;
-use actix_web::{web, HttpRequest, HttpResponse};
+use actix_web::{web, HttpRequest, HttpResponse, ResponseError};
 use serde_json::Value;
 
 use crate::event;
@@ -26,7 +28,11 @@ use crate::query::Query;
 use crate::response::{self, EventResponse};
 use crate::s3::S3;
 use crate::storage::ObjectStorage;
-use crate::utils;
+use crate::utils::header_parsing::collect_labelled;
+use crate::utils::{self, merge};
+
+const PREFIX_TAGS: &str = "x-p-tags-";
+const PREFIX_META: &str = "x-p-meta-";
 
 pub async fn query(_req: HttpRequest, json: web::Json<Value>) -> HttpResponse {
     let json = json.into_inner();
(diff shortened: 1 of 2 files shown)


server/src/handlers/event.rs
```diff
@@ -74,7 +80,16 @@ pub async fn query(_req: HttpRequest, json: web::Json<Value>) -> HttpResponse {
...
     let stream_name: String = req.match_info().get("logstream").unwrap().parse().unwrap();
-    let labels = utils::collect_labels(&req);
+
+    let tags = match collect_labelled(&req, PREFIX_TAGS, ',') {
+        Ok(tags) => HashMap::from([("p_tags".to_string(), tags)]),
+        Err(e) => return e.error_response(),
+    };
+
+    let metadata = match collect_labelled(&req, PREFIX_META, ',') {
+        Ok(metadata) => HashMap::from([("p_metadata".to_string(), metadata)]),
```
nitisht, line 90: Create a const with `,` as a value. Use that in passing as args, that would allow changing this value at a single point if needed

server/src/handlers/event.rs
```diff
@@ -74,7 +80,16 @@ pub async fn query(_req: HttpRequest, json: web::Json<Value>) -> HttpResponse {
...
 pub async fn post_event(req: HttpRequest, body: web::Json<serde_json::Value>) -> HttpResponse {
     let stream_name: String = req.match_info().get("logstream").unwrap().parse().unwrap();
-    let labels = utils::collect_labels(&req);
+
+    let tags = match collect_labelled(&req, PREFIX_TAGS, ',') {
+        Ok(tags) => HashMap::from([("p_tags".to_string(), tags)]),
+        Err(e) => return e.error_response(),
+    };
+
+    let metadata = match collect_labelled(&req, PREFIX_META, ',') {
```
nitisht, line 89: Lets have a multi levelled approach here --

`collect_tags` and `collect_meta` as top level functions, both calling  private fn `collect_from_header` with proper prefix and request.

server/src/utils.rs
```diff
@@ -17,52 +17,97 @@
...
+                return Err(ParseHeaderError::SeperatorInKey(kv_seperator));
+            }
+            if value.contains(kv_seperator) {
+                return Err(ParseHeaderError::SeperatorInValue(kv_seperator));
+            }
+
+            labels.push(format!("{}={}", key, value));
+        }
+
+        Ok(labels.join(","))
```
nitisht, line 88: I don't see the check to allow max of 5 tags and meta entries each

nitisht (changes requested):
Looking great, some comments for readability

nitisht (approved):
👍🏽

PR #111, opened 2022-09-14
Change x-p-tags to x-p-tag

### LINKED ISSUE #110: Tag header prefix should be `x-p-tag` instead of `x-p-tags` as one header adds only one tag.

trueleo:
Fixes #110

<hr>

This PR has:


nitisht (approved):
LGTM

PR #138, opened 2022-10-02
Bump aws_sdk_s3 version and provide sleep and retry impl for S3 client

### LINKED ISSUE #63: Warnings on RUST_LOG=warn
on RUST_LOG=warn there are bunch of these warnings
```
aws_smithy_client::builder] Retries require a `sleep_impl`, but none was passed into the builder. No retries will occur with the current configuration. If this was intentional, you can suppress this message with `Client::set_sleep_impl(None). Otherwise, unless you have a good reason to use the low-level service client API, consider using the `aws-config` crate to load a shared config from the environment, and construct a fluent client from that. If you need to use the low-level service client API, then pass in a sleep implementation to make timeouts and retry work.
```

trueleo:
Fixes #63 

### Description
Client is given a generic retry and sleep implementation which gets rid of warning generated by aws_sdk_s3 crate

### DIFF
--- a/server/src/s3.rs
+++ b/server/src/s3.rs
@@ -3,7 +3,9 @@ use aws_sdk_s3::error::{HeadBucketError, HeadBucketErrorKind};
 use aws_sdk_s3::model::{CommonPrefix, Delete, ObjectIdentifier};
 use aws_sdk_s3::types::{ByteStream, SdkError};
 use aws_sdk_s3::Error as AwsSdkError;
+use aws_sdk_s3::RetryConfig;
 use aws_sdk_s3::{Client, Credentials, Endpoint, Region};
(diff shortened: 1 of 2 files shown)


server/src/s3.rs
```diff
@@ -171,6 +173,8 @@ impl S3 {
             .region(options.region)
             .endpoint_resolver(options.endpoint)
             .credentials_provider(options.creds)
+            .retry_config(RetryConfig::standard().with_max_attempts(1))
```
trueleo, line 176: @nitisht  Any other suggestion on max_attempts value

nitisht: Let's change it to 5. Higher retries is generally safer.

PR #153, opened 2022-10-06
Create all temporary directories on startup

### LINKED ISSUE #74: Fails to process event if data directory has been deleted
Currently data directory is created when new stream is added but if data directory is deleted then it errors out.
 
`failed to process event because io error: No such file or directory (os error 2)`

Possible fixes:
 - Check and create data directory and all the subfolder for each stream on startup. 
 - check and create when an event is handled

trueleo:
Fixes #74.

### Description
Create all the temporary directories on program startup. Right now event writer simply creates the record file without checking or creating the directories. This responsibility should be delegated to startup sync where it creates all the directories beforehand based on metadata available. 

Startup sync now explicitly panics if create temporary directory fails ( permission issues ) or moving some old record file fails. This is because these operations are essential for running the server and it saves users from the confusion of why they  cannot post any event.
   
<hr>

This PR has:

### DIFF
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -36,7 +36,6 @@ include!(concat!(env!("OUT_DIR"), "/generated.rs"));
 use std::fs::{self, File};
 use std::io::BufReader;
 use std::panic::{catch_unwind, AssertUnwindSafe};
-use std::path::Path;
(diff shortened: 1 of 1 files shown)


nitisht (approved):
LGTM

PR #157, opened 2022-10-07
refactor: improve dockerfile using latest binary

### LINKED ISSUE #156: Improve dockerfile to use a release binary instead of building the rust binary
Current Docker build takes quite long and needs to be improved, but using a release binary directly instead of building the server in the Dockerfile

adripo:
Fixes #156

### Description

<hr>

This PR has:


nitisht:
Thank you for the PR @adripo looks great!

nitisht (approved):

PR #162, opened 2022-10-09
Add demo flag documentation in error message

### LINKED ISSUE #160: Add demo flag documentation in error message
Current error message is like

```
error: The following required arguments were not provided:
  --username <username>
  --password <password>
  --s3-endpoint-url <url>
  --s3-access-key-id <access-key>
  --s3-secret-key <secret-key>
  --s3-region <region>
  --s3-bucket-name <bucket-name>

Usage: parseable --username <username> --password <password> --s3-endpoint-url <url> --s3-access-key-id <access-key> --s3-secret-key <secret-key> --s3-region <region> --s3-bucket-name <bucket-name>

For more information try '--help'
```

We need to add the `--demo` flag option here itself. This way users know if they want they can easily experience parseable with demo mode without the needs to set up a bucket. This is how the message can look

```
You can also use the --demo flag to run Parseable with default object storage. For testing purposes only
```

sakti:
This changes add information about `--demo` flag on clap parse error if the required argument is not supplied.

Fixes #160.

### Description

Add information about demo mode to user, if they try to run the executable and not yet configured the required environment variables or flags.

<hr>

This PR has:

### DIFF
--- a/server/src/option.rs
+++ b/server/src/option.rs
@@ -53,9 +53,14 @@ where
     S: Clone + clap::Args + StorageOpt,
 {
     fn new() -> Self {
-        Config {
-            parseable: Opt::<S>::parse(),
-        }
+        let parseable = match Opt::<S>::try_parse() {
+            Ok(s) => s,
+            Err(e) => {
+                eprintln!("You can also use the --demo flag to run Parseable with default object storage. For testing purposes only");
+                e.exit();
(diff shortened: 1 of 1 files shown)


trueleo (approved):

nitisht:
Thank you for the PR @sakti !

PR #175, opened 2022-10-18
Set zip version with only deflate feature enabled

trueleo:
### Description

Set `zip` crate back to stable crates.io version with minimal features enabled. This gets rid of bzip2 as indirect dependency and transitively on bzip2-sys crate. Console releases on GitHub use `deflate` method for zip, so Parseable build step should only requires this feature.


nitisht:
@trueleo does this fix the Docker build issue we observed?

trueleo:
> @trueleo does this fix the Docker build issue we observed?

Nope, only get rids of one of the sys crate.. that issue still persists

nitisht:
> > @trueleo does this fix the Docker build issue we observed?
> 
> Nope, only gets rid of one of the `-sys` crate.. that issue still persists

okay

PR #179, opened 2022-10-22
Refactor alerts

trueleo:
### Description

Alerts are stored on metadata map which is behind and rwlock. This is fine but having to hold writer lock for alert check is not ideal. This PR refactors code around alerts implementation. Each alerts rule should be self contained with its own interior mutable state if it requires any. Interior mutablity pattern allows for alert to be resolved with shared reference. An alert can now contain different variant of rules which can be state or stateless.
<hr>

This PR has:

### DIFF
--- a/server/src/alerts.rs
+++ b/server/src/alerts.rs
@@ -113,9 +113,15 @@ impl NumericRule {
 
         let comparison = match self.operator {
             NumericOperator::EqualTo => number == &self.value,
-            // TODO: currently this is a hack, ensure checks are performed in the right way
+            NumericOperator::NotEqualTo => number != &self.value,
             NumericOperator::GreaterThan => number.as_f64().unwrap() > self.value.as_f64().unwrap(),
+            NumericOperator::GreaterThanEquals => {
+                number.as_f64().unwrap() >= self.value.as_f64().unwrap()
+            }
             NumericOperator::LessThan => number.as_f64().unwrap() < self.value.as_f64().unwrap(),
+            NumericOperator::LessThanEquals => {
+                number.as_f64().unwrap() <= self.value.as_f64().unwrap()
+            }
         };
 
         // If truthy, increment count of repeated
@@ -143,9 +149,18 @@ impl NumericRule {
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub enum NumericOperator {
+    #[serde(alias = "=")]
     EqualTo,
+    #[serde(alias = "!=")]
+    NotEqualTo,
+    #[serde(alias = ">")]
     GreaterThan,
+    #[serde(alias = ">=")]
+    GreaterThanEquals,
+    #[serde(alias = "<")]
     LessThan,
+    #[serde(alias = "<=")]
+    LessThanEquals,
 }
 
 impl Default for NumericOperator {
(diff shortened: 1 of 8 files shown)


nitisht:
Tested the PR, few observations:

1. Incorrect alert field causes server to panic. Alert config

```json
{
	"alerts": [{
		"name": "server-fail-alert1",
		"message": "server reported error status",
		"rule": {
			"field": "http_status",
            "operator": "equalTo",
			"value": 500,
			"repeats": 2,
			"within": "5m"
		},
		"targets": [{
			"name": "slack-target",
			"server_url": "https://webhook.site/70659a77-386b-4543-ab1d-98addddceeb1",
			"api_key": ""
		}]
	}]
}
```
Error:
```
thread 'actix-rt|system:0|arbiter:4' panicked at 'field exists', server/src/alerts.rs:87:51
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

server/src/alerts.rs
```diff
@@ -52,58 +54,79 @@ impl Alert {
...
     }
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
-pub enum Operator {
+pub enum NumericOperator {
     EqualTo,
     GreaterThan,
     LessThan,
```
nitisht, line 126: Let's add more rules here `NotEqualTo`, `GreaterThanEquals`, `LessThanEquals`.

nitisht (commented):
The target struct should have format like this

```
		"target_1": {
			"type": "webhook",
			"url": "https://webhook.site/70659a77-386b-4543-ab1d-98addddceeb1",
			"username": "", //optional
			"password": "" //optional
                         "token": "" //optional
		},
```

i.e. add a `type` field, that will allow adding more types later

server/src/alerts.rs
```diff
@@ -16,16 +16,18 @@
...
 #[serde(rename_all = "camelCase")]
 pub struct Alerts {
     pub alerts: Vec<Alert>,
 }
 
-#[derive(Default, Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Debug, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct Alert {
     pub name: String,
```
nitisht, line 33: As we discussed, let's add a system generated `id` here apart from alert name. While saving this to S3, let's use the `id` as the file name.

server/src/alerts.rs
```diff
@@ -52,58 +54,79 @@ impl Alert {
...
+            Rule::Numeric(rule) => rule.resolves(event),
+        }
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
-pub struct Rule {
+pub struct NumericRule {
     pub field: String,
```
nitisht, line 74: Instead of `field`, let's call it `column` - makes it simpler to understand

trueleo:
>     1. Incorrect alert field causes server to panic. Alert config

field should be checked at time of registering alert. I will add that to validation. It does not make sense to have a fallible code in there.

nitisht:
> > ```
> > 1. Incorrect alert field causes server to panic. Alert config
> > ```
> 
> field should be checked at time of registering alert. I will add that to validation. It does not make sense to have a fallible code in there.

Yes, I am not asking to validate after saving.

PR #181, opened 2022-10-23
Add path validation for cert and key

trueleo:
### Description
This PR adds validation to clap parser for `tls-cert-path` and `tls-key-path` arguments. If path is empty or does not point to a valid file then in that case program will exit with appropriate error message.

### DIFF
--- a/server/src/option.rs
+++ b/server/src/option.rs
@@ -168,11 +168,21 @@ where
     S: Clone + clap::Args + StorageOpt,
 {
     /// The location of TLS Cert file
-    #[arg(long, env = "P_TLS_CERT_PATH", value_name = "path")]
+    #[arg(
+        long,
+        env = "P_TLS_CERT_PATH",
+        value_name = "path",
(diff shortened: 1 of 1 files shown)


server/src/option.rs
```diff
@@ -251,3 +261,21 @@ where
...
+
+    pub fn file_path(s: &str) -> Result<PathBuf, String> {
+        if s.is_empty() {
+            return Err("empty path".to_owned());
+        }
+
+        let path = PathBuf::from(s);
+
+        if !path.is_file() {
+            return Err("path specified does not point to an accessible file".to_string());
```
nitisht, line 276: Print the path as well, so it is clear which path it is talking about

trueleo: Clap prints the value that was passed. 
```error: Invalid value "dir/" for '--tls-cert-path <path>': path specified does not point to an accessible file```

nitisht: oh nice

Issue #184, opened 2022-10-26
Alerts cleanup


nitisht:
There are a few more pending items before alerts are useful:

- [ ] String fields in columns.
- [ ] Internal state management (`listening` --> `triggered` --> `resolved` --> `listening`).
- [ ] Documentation.

nitisht:
fixed by #189

PR #189, opened 2022-10-29
Introduce alert state and timeouts

trueleo:
### Description

Implement alert state such that webhooks can be called on both alert trigger and alert resolution. Timeouts are also added to avoid flooding targets ( this is opt in rather than opt out because it adds overhead of state check and timer spawn mechanism ). Numeric rule and String rule are used through Consecutive* rule which takes repeat value as 1 if no repeat value is given.

ref to #184 
<hr>

This PR has:


nitisht (commented):
Tested, looking great!

Few small changes we need

1. Split the `rule` field into a `type` and `config` field. It is easy to understand for the user. We can later add more `types` like `query-based`, `status-based` etc.

```
	"rule": {
		"type": "column-based", // we'll add more types later
		"config": {
                      "column": "status",
                      "operator": ">=",
                      "value": 500,
                      "repeats": 3
		}
	},
```

2. Under targets, let's rename `timeout` --> `repeat`. `server_url` --> `url`.
3. Add validation for `repeat` field inside the target. If it is not provided, no issues, but if a junk value is sent inside `repeat` field, server should fail the set alert API.

PR #190, opened 2022-10-30
Create a Dockerfile that outputs a static binary during build

Mayon-Francis:
### Description
Can be built using:
```
docker build -f Dockerfile.static -o static-build .
```

<hr>

This PR has:


nitisht (approved):
LGTM! 🎉

Mayon-Francis:
@nitisht Should there be a `.dockerignore` to avoid copying the "target" folder when building?

Issue #146, opened 2022-10-03
[Document] How to send logs to Parseable from a Rust application


nitisht:
Parseable website has docs on how to directly send logs from Go, Python, Java, JS etc. Ref: https://www.parseable.io/docs/log-ingestion/applications/go

We need to add docs on how to directly sends logs from a Rust application. 

Please send the PR to parseablehq/parseable repo and we'll add it to the website.

nitisht:
@123vivekr would you take this up as well?

123vivekr:
Yes 👍

nitisht:
Fixed by #196

PR #197, opened 2022-11-22
Show git semver in banner

123vivekr:
Closes https://github.com/parseablehq/parseable/issues/146

### Description

Show git semver in banner

<hr>

This PR has:


nitisht:
Thanks for the PR @123vivekr . @trueleo can you take a look

trueleo:
@nitisht Looking

server/src/banner.rs
```diff
@@ -44,56 +45,104 @@ pub fn warning_line() {
     );
 }
 
+pub enum ParseableVersion {
```
trueleo, line 48: @123vivekr Can you move this inside version module below. They are not used anywhere else.

trueleo:
@nitisht Just a note,  git version is a same as build version until you add a tag like 0.0.6-*** .. So BUILD_SEMVER is the last release semver and GIT_SEMVER uses last tag ( like 0.0.6-aplha )

server/src/banner.rs
123vivekr: Ah yes. Its done

server/src/banner.rs
```diff
@@ -48,52 +48,99 @@ pub mod version {
...
 
-    pub fn current() -> Version {
-        let current = env!("VERGEN_BUILD_SEMVER");
-        semver::Version::parse(current).expect("VERGEN_BUILD_SEMVER is always valid semver")
+    pub fn current() -> ParseableVersion {
+        let build_semver = env!("VERGEN_BUILD_SEMVER");
+
+        let mut git_semver = env!("VERGEN_GIT_SEMVER");
+
+        #[allow(clippy::iter_nth_zero)]
```
trueleo, line 126: @123vivekr  I think this part of code can be replaced with 

```
if &git_semver[..1] == "v" {
    git_semver = &git_semver[1..];
}
```

PR #202, opened 2022-11-27
Change local cache directory structure

trueleo:
### Description

This PR introduces changes to local directory and removes tmp directory. All conversion is still carried out in s3 sync cycle. StorageDir is used as primary class for determining what is the data path for a logstream.

Recordbatch file is signified by extention .arrows.

This PR removes startup sync step as it is not required.

<hr>

This PR has:


nitisht:
LGTM 👍🏽

nitisht (approved):

PR #210, opened 2022-12-11
Add ingest API

123vivekr:
### Description

This API will push logs to stream based on header.

Separated out pushing log logic into `push_logs()` - avoid code redundancy

Expose `/ingest` API which push logs to a stream name based on header.
The header key is `X-P-STREAM-NAME`

<hr>

This PR has:


nitisht:
Thanks for the PR @123vivekr we'll take a look

trueleo (approved):
Thank you for the PR

PR #211, opened 2022-12-12
Add support for automatic stream creation

nitisht:
In case the user specifies x-p-stream in the header and the stream doesn't exist, server failed with stream doesn't exist error.

But in highly automated environments like kubernetes, it is important to dynamically create the stream so that Parseable is able to ingest the log data quickly without too much user involvement.

This PR has:


AdheipSingh:
@nitisht does this require change in collector ?

trueleo (approved):

PR #217, opened 2022-12-15
Add option to use local filesystem for storage

trueleo:
### Description

This PR enables Parseable server to accept additional arguments ( drive or s3) to switch between storage option.
Local filesystem behaves as an storage provider through implementation of ObjectStorage trait. All
methods of LocalFS are implemented using tokio::fs

Configuration provided on runtime is used to select a storage provider. The storage object is handled as
trait objects hence the requirement of ?Sized on some methods. Calls to S3 are replaced with appropriate methods.

<hr>

This PR has:


nitisht (approved):
LGTM, there are several improvements required, we can do those incrementally.

PR #222, opened 2022-12-22
Metadata checks

trueleo:
### Description

Parseable will check main metadata file (.parseable.json) on both storage and staging directory. If any critical differences are to be found it will cause a graceful exit. The issue can be resolved by mending the different metadata versions.


nitisht (approved):
LGTM and Tested

PR #227, opened 2022-12-27
Automatically add p_datetime if there are no suitable time columns

trueleo:
###  Description

This PR adds ability for a stream to define altered schema with a dedicated timestamp column if there is none provided in the first request. This column is called `p_datetime`

This PR has:

### DIFF
--- a/server/src/event.rs
+++ b/server/src/event.rs
@@ -47,7 +47,7 @@ type LocalWriter = Mutex<Option<StreamWriter<std::fs::File>>>;
 type LocalWriterGuard<'a> = MutexGuard<'a, Option<StreamWriter<std::fs::File>>>;
 
 const DEFAULT_TIMESTAMP_KEY: &str = "p_timestamp";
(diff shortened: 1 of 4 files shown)


server/src/event.rs
```diff
@@ -40,13 +40,15 @@ use crate::metadata;
...
 use crate::storage::StorageDir;
-use crate::utils::merge;
 
 use self::error::{EventError, StreamWriterError};
 
 type LocalWriter = Mutex<Option<StreamWriter<std::fs::File>>>;
 type LocalWriterGuard<'a> = MutexGuard<'a, Option<StreamWriter<std::fs::File>>>;
 
+const DEFAULT_TIMESTAMP_KEY: &str = "p_timestamp";
+const TIME_KEYS: &[&str] = &["time", "datetime", "timestamp"];
```
nitisht, line 50: `const TIME_KEYS: &[&str] = &["time", "datetime", "timestamp", “date”]; `

PR #272, opened 2023-02-01
Increase the request body payload size limit to 10MB (#271)

### LINKED ISSUE #271: Increase the request body payload size limit to 10MB

regularcoder:
Fixes #271.

### Description
Increase the request body payload size limit to 10MB

### DIFF
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -261,7 +261,11 @@ pub fn configure_routes(cfg: &mut web::ServiceConfig) {
             // POST "/query" ==> Get results of the SQL query passed in request body
             .service(web::resource(query_path()).route(web::post().to(handlers::event::query)))
             // POST "/ingest" ==> Post logs to given log stream based on header
-            .service(web::resource(ingest_path()).route(web::post().to(handlers::event::ingest)))
+            .service(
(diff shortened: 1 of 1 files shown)


nitisht:
@trueleo PTAL

trueleo:
@nitisht This increases the payload size to ~10mb. However just took a look and it seems like the limit is changes just for logstream service but not for ingest endpoint

trueleo (approved):

regularcoder:
Thank you @trueleo for reviewing. So @nitisht  it seems logical to also add the limit to /ingest endpoint? After all it seems to serve same purpose.

Also - please let me know if I should go ahead and fix the "warning: variables can be used directly in the `format!` string" Clippy errors that are unrelated to my change.

trueleo:
@regularcoder Yes, /ingest endpoint serves the same purpose as /logstream/{stream_name}. You can put the same JsonConfig to [/ingest](https://github.com/parseablehq/parseable/blob/main/server/src/main.rs#L264) as well

nitisht:
Thanks for the PR @regularcoder please proceed with the `ingest` changes as well. We can then merge this. You can ignore the clippy failure.

PR #290, opened 2023-02-15
Re-export deps with static feature

trueleo:


nitisht:
merged as a part of #286

Issue #311, opened 2023-02-24
Additional labels for alertmanager target types in log alert spec


nitisht:
We could also add some additional `labels` to add more context from where the alert is coming from.

https://github.com/parseablehq/parseable/blob/c15cdd1cf00888a0e33d852e866d1f5425e96b1f/server/src/alerts/target.rs#L261

Suggested additional labels: 
```
https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config

__address__ = config.parseable.address
__scheme__ = config.parseable.get_scheme()
__alerts_path__ = /api/v1/logstream/{{stream_name}}/alert

__meta_parseable_deployment_id = StorageMetadata.deployment_id
__meta_parseable_version = StorageMetadata.version
__meta_parseable_mode = StorageMetadata.mode
__meta_parseable_alert_rule_type = "column"
__meta_parseable_alert_rule_config_column = "status"
__meta_parseable_alert_rule_config_operator = "notEqualTo"
__meta_parseable_alert_rule_config_repeats = 2
__meta_parseable_alert_rule_config_value = 500
```

_Originally posted by @syepes in https://github.com/parseablehq/parseable/issues/306#issuecomment-1442498884_

nitisht:
@syepes how about clubbing the whole alert reason into a single label, like so?

```
__meta_parseable_alert_reason = "status not equal to 500, 2 times"
```

syepes:
> @syepes how about clubbing the whole alert reason into a single label, like so?
> 
> ```
> __meta_parseable_alert_reason = "status not equal to 500, 2 times"
> ```

That also would work but also having the `__meta_parseable_alert_rule_type` to know the type

nitisht:
Thanks @syepes makes sense. We looked at [templating in alertmanager](https://prometheus.io/docs/alerting/latest/notification_examples/) and now we agree on your initial proposal of separate labels for each rule field. 

This will allow users to template their rules and notifications exactly how they need it. Much more powerful this way.

syepes:
No problem, that was the initial idea to not have to re-parse the labels.

syepes:
@nitisht 
This is now working pretty good, the only missing label would be the `instance = config.parseable.address`, so we can go back to the parseable instance that generated this alert:

![CleanShot 2023-02-27 at 4 11 58](https://user-images.githubusercontent.com/333081/221602834-6b158e43-6bc7-481d-a178-1c3c06270294.png)

[...]

PR #315, opened 2023-02-27
Add version to alerts

trueleo:
### Description

* Add v1 to alerts, which is mandatory to set.
* Add symbol `=%` and `!%` for contains and not contain variant of string rule.
* Change resolved annotation for alertmanager
* Add additional labels as discussed in #311

### DIFF
--- a/server/src/alerts/target.rs
+++ b/server/src/alerts/target.rs
@@ -331,6 +331,8 @@ impl CallableTarget for AlertManager {
             AlertState::Resolved => {
                 let alert = &mut alert[0];
                 alert["labels"]["status"] = "resolved".into();
+                alert["annotations"]["reason"] =
+                    serde_json::Value::String(payload.default_resolved_string());
                 alert["endsAt"] = Utc::now()
                     .to_rfc3339_opts(chrono::SecondsFormat::Millis, true)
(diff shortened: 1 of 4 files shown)


server/src/alerts/target.rs
```diff
@@ -325,20 +327,34 @@ impl CallableTarget for AlertManager {
           }
         }]);
 
+        let alert = &mut alerts[0];
+
+        alert["labels"].as_object_mut().expect("is object").extend(
```
syepes, line 332: Small suggestion, we should not have all the rules details as the standard labels (additional_labels) as they will be visible and if just one changes it create a new alert.

We could have the more internal (metadata) labels (__meta_xx), for these kind of rule details

nitisht: Thanks @syepes . It is not clear from the docs, but are the Prometheus relabeling rules applicable for alertmanager as well?

syepes: > Thanks @syepes . It is not clear from the docs, but are the Prometheus relabeling rules applicable for alertmanager as well?

https://github.com/parseablehq/parseable/issues/311#issuecomment-1446523836

PR #320, opened 2023-02-28
Skip localfs lost+found folders

syepes:
### Description
Skips the lost+found folder when using localfs as a stream.

<hr>

This PR has:
- [X] been tested to ensure log ingestion and log query works.


trueleo (commented):
This will only fix for one of such situation. I will open a fix for this in few moment.

syepes:
replaced by https://github.com/parseablehq/parseable/pull/321

PR #397, opened 2023-05-06
Add log stream creation and set retention in Helm Chart

nitisht:


nitisht:
Merging since locally tested and works fine

PR #398, opened 2023-05-07
Retention config shouldn't need logstream to be initialized

nitisht:
### Description

It is common for users to create stream and set the retention (and be done with initial config) before sending logs to the log stream.

Accordingly we're going to add support in Helm chart to declaratively create logstream and set retention.

To allow that setup we need to ensure log stream retention can be set before log stream has events (i.e. initialized)

This PR has:


nitisht:
Required for #397

PR #399, opened 2023-05-07
Fix p_timestamp to index 0

trueleo:


nitisht:
Closing since all the changes here are also available in #402

PR #409, opened 2023-05-16
Add multi user support

trueleo:


server/src/handlers/http/rbac.rs
```diff
@@ -0,0 +1,140 @@
+use crate::{
```
nitisht, line 1: License header

server/src/handlers/http/rbac.rs
```diff
@@ -0,0 +1,140 @@
...
+    validator::{self, error::UsernameValidationError},
+};
+use actix_web::{http::header::ContentType, web, Responder};
+use http::StatusCode;
+use tokio::sync::Mutex;
+
+// async aware lock for updating storage metadata and user map atomicically
+static UPDATE_LOCK: Mutex<()> = Mutex::const_new(());
+
+pub async fn put_user(username: web::Path<String>) -> Result<impl Responder, RBACError> {
```
nitisht, line 17: Add a detailed comment here on how the API behaves - Takes username, creates the user, responds with a password. That password can't be accessed again, to change password call reset and so on.

server/src/handlers/http/rbac.rs
```diff
@@ -0,0 +1,140 @@
...
+    let username = username.into_inner();
+    // verify username
+    validator::verify_username(&username)?;
+    // get exclusive lock
+    let _ = UPDATE_LOCK.lock().await;
```
nitisht, line 22: These are superfluous comments, remove

server/src/rbac/user.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+impl User {
+    // create a new User and return self with password generated for said user.
+    pub fn create_new(username: String) -> (Self, String) {
+        let PassCode { password, hashcode } = Self::gen_new_password();
+        (
+            Self {
+                username,
+                password: hashcode,
```
nitisht, line 27: why is password set to hashcode and a separate password returned later?

server/src/rbac/user.rs
```diff
@@ -0,0 +1,87 @@
...
+        Argon2::default()
+            .verify_password(password.as_bytes(), &parsed_hash)
+            .is_ok()
+    }
+
+    // gen new password
+    pub fn gen_new_password() -> PassCode {
+        let password = Alphanumeric.sample_string(&mut rand::thread_rng(), 16);
+        let hashcode = gen_hash(&password);
+        PassCode { password, hashcode }
```
nitisht, line 48: if the function is called `gen_new_password` only generate the password  here and create hash in separate step.

nitisht (commented):
few comments

server/src/migration/metadata_migration.rs
```diff
@@ -0,0 +1,31 @@
...
+
+use serde_json::Value;
+
+pub fn v1_v2(mut storage_metadata: serde_json::Value) -> Value {
+    let metadata = storage_metadata.as_object_mut().unwrap();
+    *metadata.get_mut("version").unwrap() = Value::String("v2".to_string());
+    metadata.remove("user");
+    metadata.remove("stream");
+    metadata.insert("users".to_string(), Value::Array(vec![]));
+    metadata.insert("streams".to_string(), Value::Array(vec![]));
```
nitisht, line 29: Also need to convert `role` to `roles` - since that is an array.

trueleo: Changed to User attribute to "roles". No migration step needed for this tho

PR #411, opened 2023-05-17
skip secret option added

### LINKED ISSUE #408: Skip secrets option
While setting parseable server, I got to know how passing a secrets can be frustrating for beginner user. We can add a option to skip secrets. So user can start server with ease. I would like to take this issue up.

Sajiyah-Salat:
This pr is to give a smooth installation experience to user by keeping secrets optional. In this way user dont have to worry about secrets and they can start server easily. 

Fixes #408 

This PR has:


nitisht:
Closing this for now, because we're looking for another approach.

PR #419, opened 2023-05-22
Deleting a logstream should delete its stats. 

### LINKED ISSUE #412: bug: after deleting a log stream, still able to get the stats for that logstream
Steps 

1. Create a log stream.
2. Ingest events to the log stream.
3. Get stats of the log stream.
4. Delete the log stream.
5. Call list stream API to see the stream is actually deleted.
6. Get stats of the log stream --> expectation is that API fails with 404, but currently it returns the last recorded stats of the stream.

davidrhp:
Fixes #412 .

### Description
The intention is to fix the bug mentioned in the issue, where get_stats reported the last recorded data of a logstream, even though it has already been deleted. Since the logstream has already been deleted, no stats data should be available.

#### Returning a 404
Currently, the `stats::get_current_stats` also returns Stats for logstreams that do not exist and fills the struct with 0 values. This leads to callers not actually knowing, whether stats for this stream actually exist. While the signature actually uses an `Option<Stats>`, the `None` variant is only returned, if the underlying prometheus function returns an `Err`. According to the docs, this should only happen, if the number of labels don't match.

Since the underlying `MetricVec`s containing the stats, don't seem to have a method to check if the passed in labels already exist, it seems to be better to check at the caller's side whether a stream exists using `metadata::STREAM_INFO.stream_exists(..)`. If it does, check the stored metrics (stats), if not, return a 404.

#### Deleting stats when a stream is deleted
I added a function in the `stats` module to delete the metrics for the deleted logstream, which is then called in the delete logstream handler.

Since the `prometheus::vec::MetricVec::get_metric_with_label_values` and `prometheus::vec::MetricVec::remove_label_values` explicitly warn of passing multiple labels into the respective function, I extracted the label creation into their own functions, so both the `stats::get_current_stats` and the new `stats::delete_stats` always use the same labels.

**Warning**
Previously, the `get_stats` handler returned a `200`, even for logstreams that have never been created. Now, it returns a `404` for logstreams that do not exist.

<hr>

This PR has:

### DIFF
--- a/server/src/handlers/http/logstream.rs
+++ b/server/src/handlers/http/logstream.rs
@@ -46,6 +46,9 @@ pub async fn delete(req: HttpRequest) -> Result<impl Responder, StreamError> {
     objectstore.delete_stream(&stream_name).await?;
     metadata::STREAM_INFO.delete_stream(&stream_name);
     event::STREAM_WRITERS.delete_stream(&stream_name);
+    stats::delete_stats(&stream_name, "json").unwrap_or_else(|e| {
+        log::warn!("failed to delete stats for stream {}: {:?}", stream_name, e)
+    });
 
     let stream_dir = StorageDir::new(&stream_name);
     if fs::remove_dir_all(&stream_dir.data_path).is_err() {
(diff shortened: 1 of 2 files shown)


davidrhp:
@trueleo I am ready with this PR, however, I am not sure about this point:

"[] added documentation for new or modified features or behaviors."

This PR will change the Stream Storage Stats API by returning 404s for logstreams that do not exists. Previously, it has returned a 200, with zeros for the respective metrics. 

I could not find a section in the docs where this behavior was documented. If it is not, I would mark this PR as ready.

nitisht:
Thanks for the PR @davidrhp . 

> "[] added documentation for new or modified features or behaviors."

The documentation is unfortunately not yet open source and you're correct in that it doesn't cover the response. So I think we're fine on this point. Please mark this ready and we can merge it soon

Issue #250, opened 2023-01-09
feature: multi user support and access control based on RBAC


nitisht:
This includes following items

- [x] Multi user support --> Done with PR #409 
- [x] Authorisation mechanism based on roles setting per user --> Done with PR #417 
- [x] Integration tests in Quest --> PR https://github.com/parseablehq/quest/pull/21
- [x] Documentation
- [x] High CPU usage in hashing. Need to change to another hashing approach. --> #425 
- [x] List user roles API --> `GET {{endpoint}}/api/v1/user/{{username}}/role` --> #425 
- [x] Label based access for reader privilege --> #429

nitisht:
Feature is complete now.

PR #425, opened 2023-05-27
Refactor auth flow 

trueleo:
Part of #250  

### Description
 - Authentication and Authorization are merged into single middleware.
 - Added AuthMap for better permission lookup
 - Permissions are loaded lazily into AuthMap once user is authenticated


trueleo:

server/src/handlers/http/middleware.rs
```diff
@@ -75,25 +76,33 @@ where
...
+    fn call(&self, mut req: ServiceRequest) -> Self::Future {
+        let creds = req.extract::<BasicAuth>().into_inner();
+        let creds: Result<(String, String), Error> = creds.map_err(Into::into).and_then(|creds| {
+            let username = creds.user_id().trim().to_owned();
+            if let Some(password) = creds.password() {
+                Ok((username, password.trim().to_owned()))
+            } else {
+                Err(ErrorBadRequest("Password is required in basic auth"))
+            }
+        });
```
nitisht, line 88: need comments in this part - what and why are we doing this

server/src/handlers/http/middleware.rs
```diff
@@ -75,25 +76,33 @@ where
...
-        let extensions = req.extensions();
-        let username = extensions
-            .get::<String>()
-            .expect("authentication layer verified username");
-        let is_auth = Users.check_permission(username, self.action, stream);
-        drop(extensions);
+        let is_auth = creds.map(|creds| {
+            let (username, password) = creds;
+            Users.authenticate(username, password, self.action, stream)
+        });
```
nitisht, line 97: what is `is_auth`?

server/src/rbac.rs
```diff
@@ -16,136 +16,139 @@
...
 
+pub mod auth;
 pub mod role;
 pub mod user;
 
-pub static USER_AUTHENTICATION_MAP: OnceCell<RwLock<UserMap>> = OnceCell::new();
-pub static USER_AUTHORIZATION_MAP: OnceCell<RwLock<UserPermMap>> = OnceCell::new();
+pub static USER_MAP: OnceCell<RwLock<UserMap>> = OnceCell::new();
+pub static AUTH_MAP: OnceCell<RwLock<AuthMap>> = OnceCell::new();
+
```
nitisht, line 37: What is the purpose of two different maps here?

server/src/rbac/auth.rs
```diff
@@ -0,0 +1,49 @@
+use std::collections::HashMap;
```
nitisht, line 1: License

server/src/handlers/http/middleware.rs
trueleo: The type is result of authentication and authorization. Type of creds is Result<> which i can only return in the async block down below. So have to `.map` for calling authenticate

`Users.authenticate` does both authentication and authorization.

server/src/rbac.rs
trueleo: If any changes made to user ( password change, role update, etc.. ) it is updated in memory in the Usermap `Hashmap<username, User>`. The User contains password hash and role information. While `AuthMap` is for quick authentication and authorization, it is HashMap<(user, pass), Vec<Permission>>. Any changes made to User will reset the AuthMap and upon authentication user perms is reloaded into this map

nitisht (approved):
LGTM & Tested

PR #426, opened 2023-05-30
Unify RBAC flow in a single crate

nitisht:
This commit combines all the RBAC specific code in a single
crate. This improves readability.

Also added more relevant comments


nitisht:
recheck

nitisht:
@trueleo PTAL

trueleo:
@nitisht Looks all good to me but one small thing is that we should keep `src/rbac.rs` on top level instead of moving it mod.rs  

https://doc.rust-lang.org/book/ch07-05-separating-modules-into-different-files.html#alternate-file-paths

trueleo (approved):

Issue #430, opened 2023-06-04
Slow performace when using ORDER BY


Ryiski:
I've noticed queries are running extremely slow when I try and `ORDER BY` on the logs

You can give it a test on Postman

Slow query:
```json
{
    "query": "SELECT * FROM frontend ORDER BY p_timestamp LIMIT 30 OFFSET 0",
    "startTime": "2022-03-01T00:00:00Z",
    "endTime": "2023-06-03T23:59:59.999Z"
}
```

Fast query:
```json
{
    "query": "SELECT * FROM frontend LIMIT 30 OFFSET 0",
    "startTime": "2022-03-01T00:00:00Z",
    "endTime": "2023-06-03T23:59:59.999Z"
}
```

nitisht:
Thanks @Ryiski we’ll take a look.

nitisht:
Will be fixed by #443

PR #434, opened 2023-06-09
Test Sort Order

trueleo:
Fixes #XXXX.

### Description

<hr>

This PR has:


nitisht:
Let's prioritise this one as well. This looks like very useful in improving performance. relevant to #430

trueleo:
Closing this as i could not verify effects of this. Will try to add this again after #438

PR #443, opened 2023-06-24
Use sort ordering on timestamp array

### LINKED ISSUE #418: v0.4.4 Slow performance
For some reason, v0.4.4 seems to be slow.

When I create a log, it takes time to show in the stream list

Here is a visual example of v0.4.4

https://github.com/parseablehq/parseable/assets/47484379/4c257d6f-3a06-447a-b332-6a751338b4b1

Here is a visual example of v0.4.3 which is much faster

https://github.com/parseablehq/parseable/assets/47484379/85b14dc4-6adf-448a-b18b-d81574161695

### LINKED ISSUE #430: Slow performace when using ORDER BY

trueleo:
Fixes #430.

### Description

Write timestamp sortedness metadata to parquet and provide external sort information to datafusion. This way the SortExec can be avoided in execution plan with most queries which use `order by p_timestamp`. 

### Example 
`explain select p_timestamp from {{stream_name}} order by p_timestamp asc`

In physical plan it is visible that SortExec is eliminated as output_ordering is pushed to ParquetExec node 
```
"plan": "SortPreservingMergeExec: [p_timestamp@0 ASC NULLS LAST]
  ParquetExec: file_groups={4 groups: [.....]}, projection=[p_timestamp], output_ordering=[p_timestamp@0 ASC NULLS LAST]",
```

### Note:

This is still not the most optimized version of this query as SortPreservingExec is not really needed here. The issue here is that the Datafusion is not aware that the partitions / files are non overlapping when considering timestamp 

Also if the target partition limit is crossed then datafusion again adds SortExec to physical plan.

<hr>

This PR has:

### DIFF
--- a/server/src/storage/localfs.rs
+++ b/server/src/storage/localfs.rs
@@ -24,25 +24,14 @@ use std::{
 
 use async_trait::async_trait;
 use bytes::Bytes;
-use datafusion::{arrow::datatypes::Schema, prelude::col};
-use datafusion::{
-    datasource::{
-        file_format::parquet::ParquetFormat,
-        listing::{ListingOptions, ListingTable, ListingTableConfig, ListingTableUrl},
-    },
-    error::DataFusionError,
-    execution::runtime_env::RuntimeConfig,
-};
+use datafusion::{datasource::listing::ListingTableUrl, execution::runtime_env::RuntimeConfig};
 use fs_extra::file::{move_file, CopyOptions};
 use futures::{stream::FuturesUnordered, TryStreamExt};
 use relative_path::RelativePath;
 use tokio::fs::{self, DirEntry};
 use tokio_stream::wrappers::ReadDirStream;
 
-use crate::{
-    event::DEFAULT_TIMESTAMP_KEY,
-    metrics::storage::{localfs::REQUEST_RESPONSE_TIME, StorageMetrics},
-};
+use crate::metrics::storage::{localfs::REQUEST_RESPONSE_TIME, StorageMetrics};
 use crate::{option::validation, utils::validate_path_is_writeable};
 
 use super::{object_storage, LogStream, ObjectStorage, ObjectStorageError, ObjectStorageProvider};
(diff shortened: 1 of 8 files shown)


server/src/storage/localfs.rs
```diff
@@ -222,12 +225,12 @@ impl ObjectStorage for LocalFS {
...
         let listing_options = ListingOptions {
             file_extension: ".parquet".to_string(),
-            file_sort_order: Vec::new(),
+            file_sort_order: vec![vec![col(DEFAULT_TIMESTAMP_KEY).sort(true, false)]],
             infinite_source: false,
             format: Arc::new(file_format),
             table_partition_cols: vec![],
             collect_stat: true,
-            target_partitions: 1,
+            target_partitions: 32,
```
nitisht, line 233: What the significance of changing this field `target_partitions`

server/src/storage/s3.rs
```diff
@@ -474,12 +476,12 @@ impl ObjectStorage for S3 {
...
         let listing_options = ListingOptions {
             file_extension: ".parquet".to_string(),
-            file_sort_order: Vec::default(),
+            file_sort_order: vec![vec![col(DEFAULT_TIMESTAMP_KEY).sort(true, false)]],
             infinite_source: false,
             format: Arc::new(file_format),
             table_partition_cols: vec![],
             collect_stat: true,
-            target_partitions: 1,
+            target_partitions: 32,
```
nitisht, line 484: This section is repeated for local and s3 mode. Can we move it to the common abstraction?

trueleo: Yes, this needs refactoring

server/src/storage/localfs.rs
trueleo: Roughly the partition here is number of parallel streams that is generated by datafusion during execution. Having this 1 was causing all files to be grouped in one partition and datafusion is unable to use external sort information for files in a group as it cannot infer order between grouped files and if they are overlapping in time range or not.

trueleo: So at max datafusion will hold 32 streams to calculate the output and merge them back using  (SortPreservingMerge)[https://docs.rs/datafusion/latest/datafusion/physical_plan/sorts/sort_preserving_merge/struct.SortPreservingMergeExec.html]

nitisht:
> Fixes https://github.com/parseablehq/parseable/issues/418.

Shouldn't this be #430? #418 is related to staging query

PR #448, opened 2023-06-30
Query staging data

### LINKED ISSUE #445: Query staging data by reading .arrows files
Tracking issue for local query.

Todo: 
- ~Add .arrows format by implementing [FileFormat](https://docs.rs/datafusion/latest/datafusion/datasource/file_format/trait.FileFormat.html). Datafusion only supports IPC file format not the streaming format which we use.~
- [x] Add arrow partition streams to table provider
- [x] While reading compact record batches to save space. 
- [x] Create table provider that joins local and remote query together in right order. 
- [ ] ( Optional ) Support projection pushdown.
- [ ]  Add atomic file references so that no arrow files are removed during query.

trueleo:
Fixes #445.

### Description
 Implements and makes use of a custom query table provider which combines  `ListingTable` and `MemTable` for querying remote data and staging data together.  

[UnionExec](https://docs.rs/datafusion/latest/datafusion/physical_plan/union/struct.UnionExec.html) is used to combine two inputs together. This can be switched out for SortPreservingMergeExec if output is interleaving staging and remote data together on something like `select * from table`

<hr>

This PR has:

### DIFF
--- a/server/src/event/writer.rs
+++ b/server/src/event/writer.rs
@@ -25,9 +25,12 @@ use std::{
     sync::{Arc, Mutex, RwLock},
 };
 
+use crate::utils;
+
 use self::{errors::StreamWriterError, file_writer::FileWriter, mem_writer::MemWriter};
-use arrow_array::RecordBatch;
+use arrow_array::{RecordBatch, TimestampMillisecondArray};
 use arrow_schema::Schema;
+use chrono::Utc;
 use derive_more::{Deref, DerefMut};
 use once_cell::sync::Lazy;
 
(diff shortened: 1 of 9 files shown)


server/src/event/writer.rs
```diff
@@ -18,21 +18,52 @@
...
+use arrow_schema::Schema;
+use chrono::Utc;
 use derive_more::{Deref, DerefMut};
 use once_cell::sync::Lazy;
 
 pub static STREAM_WRITERS: Lazy<WriterTable> = Lazy::new(WriterTable::default);
 
+#[derive(Default)]
+pub struct Writer {
+    pub mem: MemWriter<1024>,
```
nitisht, line 41: Change this to 16384

nitisht (changes requested):

PR #469, opened 2023-08-07
Query param for attaching fields to response json

trueleo:
### Description

This PR adds query parameters to format response output. Mainly duplicating existing fill null option to query param and adding a `withFields` param options.

Attaching schema to output can benefit cases where client wants to display table but the table API wants all table columns to be defined first. 

#### Example 
```
{{endpoint}}/api/v1/query?fields=true
```
```
"query": "select min(p_timestamp), max(p_timestamp) from {{stream_name}}",
....
```
```
{
    "fields": [
        "MIN(app.p_timestamp)",
        "MAX(app.p_timestamp)"
    ],
    "records": [
        {
            "MAX(app.p_timestamp)": "2023-08-07T05:55:59.738",
            "MIN(app.p_timestamp)": "2023-08-07T05:55:20.335"
        }
    ]
}
``` 

<hr>

This PR has:

### DIFF
--- a/server/src/handlers/http/query.rs
+++ b/server/src/handlers/http/query.rs
@@ -44,7 +44,7 @@ pub async fn query(
     let time = Instant::now();
 
     // format output json to include field names
-    let with_fields = params.get("withFields").cloned().unwrap_or(false);
+    let with_fields = params.get("fields").cloned().unwrap_or(false);
     // Fill missing columns with null
     let fill_null = params
         .get("fillNull")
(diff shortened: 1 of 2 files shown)


server/src/handlers/http/query.rs
```diff
@@ -36,22 +37,37 @@ use crate::rbac::role::{Action, Permission};
...
 
-pub async fn query(query: Query) -> Result<impl Responder, QueryError> {
+pub async fn query(
+    query: Query,
+    web::Query(params): web::Query<HashMap<String, bool>>,
+) -> Result<impl Responder, QueryError> {
     let time = Instant::now();
 
+    // format output json to include field names
+    let with_fields = params.get("withFields").cloned().unwrap_or(false);
```
nitisht, line 47: lets rename `withFields` --> `fields`

nitisht (approved):
small change, other wise LGTM

PR #488, opened 2023-09-07
add api route for llm query

aldrinjenson:
### Description

- Add API route for LLM query
- read from OPENAI_API_KEY in env var

<hr>

This PR has:

### DIFF
--- a/server/src/handlers/http.rs
+++ b/server/src/handlers/http.rs
@@ -231,8 +231,13 @@ pub fn configure_routes(cfg: &mut web::ServiceConfig) {
                 ),
         );
 
-    let llm_query_api =
-        web::scope("/llm").service(web::resource("").route(web::post().to(llm::make_llm_request).authorize(Action::Query)));
+    let llm_query_api = web::scope("/llm").service(
+        web::resource("").route(
+            web::post()
+                .to(llm::make_llm_request)
+                .authorize(Action::Query),
+        ),
+    );
 
     // Deny request if username is same as the env variable P_USERNAME.
     cfg.service(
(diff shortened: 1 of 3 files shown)


aldrinjenson:

server/src/handlers/http.rs
```diff
@@ -229,6 +230,10 @@ pub fn configure_routes(cfg: &mut web::ServiceConfig) {
                         .wrap(DisAllowRootUser),
                 ),
         );
+
+    let llm_query_api =
+        web::scope("/llm").service(web::resource("").route(web::post().to(llm::make_llm_request)));
```
nitisht, line 235: do we have permissions set up for this API - who can call this API?

aldrinjenson: Hi, missed permissions I'm afraid. 
Rn anyone is able to call the API.
Will modify this to need the authorization base64 string being passed in header.

trueleo (approved):

server/src/option.rs
```diff
@@ -351,6 +356,24 @@ impl Server {
...
+                    .value_name("BOOL")
+                    .required(false)
+                    .default_value("true")
+                    .value_parser(value_parser!(bool))
+                    .help("Disable/Enable sending anonymous user data"),
+            )
+            .arg(
+                Arg::new(Self::OPEN_AI_KEY)
+                    .long(Self::OPEN_AI_KEY)
+                    .env("OPENAI_API_KEY")
```
nitisht, line 372: This should be `P_OPENAI_API_KEY`, so we're consistent with the format we've used till now.

aldrinjenson: Resolved in https://github.com/parseablehq/parseable/pull/492 👍

PR #493, opened 2023-09-11
Move is llm active API to about section API

aldrinjenson:
Related to https://github.com/parseablehq/console/pull/133

### DIFF
--- a/server/src/banner.rs
+++ b/server/src/banner.rs
@@ -59,16 +59,23 @@ fn status_info(config: &Config, scheme: &str, id: Uid) {
         credentials = "\"Using default creds admin, admin. Please set credentials with P_USERNAME and P_PASSWORD.\"".red().to_string();
     }
 
+    let llm_status = match &config.parseable.open_ai_key {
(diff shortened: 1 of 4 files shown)


server/src/banner.rs
```diff
@@ -59,16 +59,23 @@ fn status_info(config: &Config, scheme: &str, id: Uid) {
...
+    };
+
     eprintln!(
         "
     {}
         URL:                {}
         Credentials:        {}
-        Deployment UID:     \"{}\"",
+        Deployment UID:     \"{}\"
+        LLM Status:          {}",
```
nitisht, line 73: Let's print the Status with double quotes as we do for all other fields

```
    Server:
        URL:                "http://0.0.0.0:8000"
        Credentials:        "Using default creds admin, admin. Please set credentials with P_USERNAME and P_PASSWORD."
        Deployment UID:     "01H8H78H9YYPNZQ3CTJPYAA3ZY"
        LLM Status:          "Not Configured"
```

aldrinjenson: Yes. Quotes Added.

PR #496, opened 2023-09-12
WIP: Do not Merge: Release workflows

aldrinjenson:
Fixes #XXXX.

### Description

<hr>

This PR has:


.github/workflows/release_mac.yml
```diff
@@ -0,0 +1,50 @@
+name: Mac
+
+on: [push]
```
nitisht, line 3: just `push` wont work AFAICS

aldrinjenson: Yes. trying to identify the correct event to build on a new commit to this pr

aldrinjenson:
Moving to a separate PR with just the workflow files  and removing all commits used while testing.

PR #514, opened 2023-09-20
create checksum.txt from all uploaded binaries

aldrinjenson:
- modify workflow to do renaming before uploading 
- uses Unix sed to remove folder names from checksum.


aldrinjenson:
https://github.com/aldrinjenson/parseable/actions/runs/6252234590

Here is a sample workfow run and the artifacts+checksum generated.

PR #518, opened 2023-09-26
Fetch ui version from Cargo toml

### LINKED ISSUE #515: Add console version details in about api
Currently it is not possible to ascertain console version at all, while running a Parseable instance. We can add console version in the about api response, it will be helpful for the users

trueleo:
Fixes #515 .

### Description

Add `uiVersion` to about handler response. This value is exported at build time.

<hr>

This PR has:

### DIFF
--- a/server/src/handlers/http/about.rs
+++ b/server/src/handlers/http/about.rs
@@ -44,9 +44,11 @@ pub async fn about() -> Json<serde_json::Value> {
     let store = CONFIG.storage().get_endpoint();
     let is_llm_active = &CONFIG.parseable.open_ai_key.is_some();
     let llm_provider = is_llm_active.then_some("OpenAI");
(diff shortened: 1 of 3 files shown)


server/src/handlers/http/about.rs
```diff
@@ -44,9 +44,11 @@ pub async fn about() -> Json<serde_json::Value> {
     let store = CONFIG.storage().get_endpoint();
     let is_llm_active = &CONFIG.parseable.open_ai_key.is_some();
     let llm_provider = is_llm_active.then_some("OpenAI");
+    let ui_version = option_env!("UI_VERSION");
 
     Json(json!({
         "version": current_version,
+        "uiVersion": ui_version,
```
nitisht, line 51: if `UI_VERSION` is not set, let us send `Development Version` as the response in the API.

PR #528, opened 2023-10-05
API Routes for Modules

### LINKED ISSUE #512: API for anomaly detection config
- Let's say Parseable does anomaly detection.
- let's scope it under stream

So from the UI(console) the user can set up different parameters to watch out for anomaly 

eg: to train a model for z-score, the user could pass in a JSON like the following:
```js
{
  "start_time": "2023-09-20T08:41:11.398Z",  // training interval start timestamp
  "end_time": "2023-09-20T13:41:11.398Z", //  // training interval end timestamp
  "stream": "frontend",
  "interval_window": '1m', // string with number and a unit of (h,m,s)
  "algorithm": "z-score" // the model to be used
}
```

We could save this in parseable server and then an anomaly detection server or plugin could just get this from parseable whenever it requires (eg: say when it restarts).
Hence, an API for config eg: /api/v1/anomaly/config (both GET and PUT methods to get and update configs) would be great.
As soon as the config is set, parseable can then call the anomaly detection server to start training based on the this new config (or some other way based on a good architecture)

We could also have another api to list out models/algorithms etc in future maybe.

P.S. This plan is in early stages and we can ideate and discuss on better solutions too.

aldrinjenson:
Fixes #512 

### Description

Adds APIs for connecting Parseable with extendable modules

- registrations saved in .parseable.json
- configs saved in stream-name/.stream.json
- other routes are directly proxied to module path

## Final Routes:

GET /api/v1/modules/  : get list of all module names
PUT  /api/v1/modules/{module} : update registration for given module
DEL  /api/v1/modules/{module} : delete registration for given module

PUT /api/v1/{module}/config/{logstream} : Update config for a given module and proxy to module config path
GET /api/v1/{module}/config/{logstream} :  Get config  object for a given stream. (Called by module)

ALL  /api/v1/{module}/* : directly proxied to module by parseable

<hr>

This PR has:


nitisht:
@aldrinjenson can you check the Deepsource / Rust issue here

server/src/handlers/http/external.rs
```diff
@@ -0,0 +1,235 @@
...
+        let module_registry = registry.read().unwrap();
+        let registration = module_registry
+            .get(&name)
+            .ok_or_else(|| ModuleError::ModuleNotFound(name.clone()))?;
+
+        let module_path = registration.get_module_path(&path, method);
+        let module_path = module_path.unwrap_or_else(|| "".to_string());
+
+        if module_path.is_empty() {
+            return Ok(HttpResponse::NotFound().finish());
```
aldrinjenson, line 152: @trueleo please see if this approach is fine.

trueleo: Refactored this part.  please take a look

trueleo (changes requested):

server/src/handlers/http/external.rs
aldrinjenson: This seems much cleaner!💯

server/src/handlers/http/external.rs
```diff
@@ -0,0 +1,250 @@
+use std::sync::RwLock;
```
nitisht, line 1: license

PR #534, opened 2023-10-19
Module API with test cases

aldrinjenson:
Based on #528 , 
but modified to work with test cases from Quest


nitisht:
Closing this for now, we'll take this in later.

PR #561, opened 2023-11-28
Catalog

trueleo:
Fixes #XXXX.

### Description

<hr>

This PR has:


server/src/catalog.rs
```diff
@@ -0,0 +1,187 @@
...
+use self::{column::Column, snapshot::ManifestItem};
+
+pub mod column;
+pub mod manifest;
+pub mod snapshot;
+
+pub use manifest::create_from_parquet_file;
+
+pub trait Snapshot {
+    fn id(&self) -> Ulid;
```
nitisht, line 40: Why is `id` needed here?

server/src/catalog.rs
```diff
@@ -0,0 +1,187 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+use std::{collections::HashMap, sync::Arc};
```
nitisht, line 19: Add a detailed comment here on what is catalog, snapshot and manifest file.

server/src/query.rs
```diff
@@ -112,56 +79,39 @@ impl Query {
...
-            state = state.add_optimizer_rule(Arc::new(filter))
-        }
+
+        let mut config = SessionConfig::default()
+            .with_parquet_pruning(true)
+            .with_prefer_existing_sort(true)
+            .with_round_robin_repartition(true)
+            .with_repartition_aggregations(false);
+
+        config.options_mut().execution.parquet.bloom_filter_enabled = true;
```
nitisht, line 89: We decided to disable bloom filters, right?

nitisht (changes requested):
This is not able to query any data that was ingested on older versions.

server/src/catalog.rs
trueleo: Id would be needed for optimistic update of snapshot in a multi node scenario. This is not in current scope so i am removing it for now.

PR #565, opened 2023-12-08
Fix broken link README.md

artraf:
### Description

Fix a broken link in README.md

This PR has:


nitisht:
Thanks for fixing @artraf

PR #566, opened 2023-12-09
Use custom table provider to reuse statistics.

trueleo:
Fixes #XXXX.

### Description
Make Table provider implementation directly create physical plan from manifest file entries. File level statistics are aggregated into table level statistics which is available for count optimization. 

Note: Aggregate Statistics will only take effect for time range which lie exactly on minute boundary. Otherwise they are treated as partial filters and plan changes, hence will not use statistics anymore.
<hr>

This PR has:

### DIFF
--- a/server/src/query/stream_schema_provider.rs
+++ b/server/src/query/stream_schema_provider.rs
@@ -16,33 +16,35 @@
  *
  */
 
-use std::{any::Any, collections::HashMap, ops::Bound, pin::Pin, sync::Arc};
+use std::{any::Any, collections::HashMap, ops::Bound, sync::Arc};
 
-use arrow_schema::SchemaRef;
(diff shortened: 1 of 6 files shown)


server/src/query/stream_schema_provider.rs
```diff
@@ -150,6 +150,10 @@ impl StandardTableProvider {
             }
         }
 
+        if manifest_files.is_empty() {
+            return Ok(None);
+        }
```
theteachr, line 155: We can return early, I think. If we move this to 130.

nitisht: is this addressed?

trueleo: There are calls to retain and truncate after 130. I think it is better to perform this check after all mutations to list have happened

PR #567, opened 2023-12-11
Add support for multiple column names

### LINKED ISSUE #564: Message format feature not replacing placeholders with correct value
Was trying to use the message format feature described in: 
https://www.parseable.io/docs/alerts#message and [#331](https://github.com/parseablehq/parseable/issues/331)

i can create an alert with this message property value: `"message": "Error: {label} {message}",`

But when I trigger an alert, the message sent to the Slack target webhook is still: 
"message": "Error: {label} {message}",

The placeholders and not replaced by the corresponding values.

Also, the documentation states: 

> The {**status_code**} is a placeholder that will be replaced with the value of the **_status_** column in the event that triggered the alert.

For the sake of clarity, should we read instead: [...] value of the **_status_code_** column [...] ?

Thanks!

_ref: Slack conversation with @nitisht_ 

**Installed version**

Commit:41029fd
Version:v0.7.1
UI Version:v0.3.4

nitisht:
Fixes #564.

### Description

Add support for multiple column names placeholders in the message field
of an alert. This is useful to set more useful alert messages for the targets.

<hr>

This PR has:

### DIFF
--- a/server/src/alerts/mod.rs
+++ b/server/src/alerts/mod.rs
@@ -135,40 +135,39 @@ pub struct Message {
 
 impl Message {
     // checks if message (with a column name) is valid (i.e. the column name is present in the schema)
-    pub fn valid(&self, schema: &Schema, column: Option<&str>) -> bool {
-        if let Some(col) = column {
-            return get_field(&schema.fields, col).is_some();
-        }
-        true
+    pub fn valid(&self, schema: &Schema, column: &str) -> bool {
+        return get_field(&schema.fields, column).is_some();
     }
 
-    pub fn extract_column_name(&self) -> Option<&str> {
+    pub fn extract_column_names(&self) -> Option<Vec<&str>> {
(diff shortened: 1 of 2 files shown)


server/src/alerts/mod.rs
```diff
@@ -135,40 +135,39 @@ pub struct Message {
...
-        if let Some(col) = column {
-            return get_field(&schema.fields, col).is_some();
-        }
-        true
+    pub fn valid(&self, schema: &Schema, column: &str) -> bool {
+        return get_field(&schema.fields, column).is_some();
     }
 
-    pub fn extract_column_name(&self) -> Option<&str> {
+    pub fn extract_column_names(&self) -> Option<Vec<&str>> {
```
theteachr, line 142: We can skip wrapping the `Vec` in an `Option`. I'll push the update.

trueleo (approved):

PR #570, opened 2023-12-14
Transform with table reference

### LINKED ISSUE #572: Query execution fails for any query that uses p_timestamp
### Description

**branch**: main

Running this specific query causes a datafusion runtime error

```sql
select p_timestamp from table;
```
``` 
Error during planning: required columns can't push down, columns: {Column { relation: Some(Bare { table: "integrity" }), name: "p_timestamp" }, Column { relation: None, name: "p_timestamp" }}
```

### Expected behavior
Query should works without any issue.

trueleo:
Fixes #572.

### Description
While transforming logical plan to add external time field, the column name that was provided is unqualified ( without table reference ).
this caused mismatch in column name in datafusion query runtime.
``` 
Error during planning: required columns can't push down, columns: {Column { relation: Some(Bare { table: "integrity" }), name: "p_timestamp" }, Column { relation: None, name: "p_timestamp" }}
```

In this PR the filter expr `start_time_filter` and `end_time_filter` for logical plan is created with table reference so that push down predicate can work properly.   

<hr>

This PR has:

### DIFF
--- a/server/src/query.rs
+++ b/server/src/query.rs
@@ -21,8 +21,8 @@ mod listing_table_builder;
 mod stream_schema_provider;
 mod table_provider;
 
-use chrono::TimeZone;
 use chrono::{DateTime, Utc};
+use chrono::{NaiveDateTime, TimeZone};
 use datafusion::arrow::record_batch::RecordBatch;
 
 use datafusion::common::tree_node::{Transformed, TreeNode, TreeNodeVisitor, VisitRecursion};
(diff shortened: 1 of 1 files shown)


trueleo:
Closes #572

nitisht:
> Let's add few queries in Quest so we catch such failures later.

https://github.com/parseablehq/quest/pull/48

nitisht (approved):
LGTM and tested locally

PR #571, opened 2023-12-14
Match table statistics with table schema

trueleo:
Fixes #XXXX.

### Description
Make Column Statistics line up with table schema as internally this is accessed via indexing

<hr>

This PR has:

### DIFF
--- a/server/src/query/stream_schema_provider.rs
+++ b/server/src/query/stream_schema_provider.rs
@@ -18,7 +18,7 @@
 
 use std::{any::Any, collections::HashMap, ops::Bound, sync::Arc};
 
-use arrow_schema::{DataType, Schema, SchemaRef, SortOptions};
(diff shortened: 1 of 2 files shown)


server/src/query/stream_schema_provider.rs
```diff
@@ -236,57 +236,55 @@ fn partitioned_files(
...
-            .get(field.name())
-            .and_then(|stats| stats.as_ref())
-        else {
-            statistics.push(datafusion::common::ColumnStatistics::default());
-            break;
-        };
+    let statistics = table_schema
+        .fields()
+        .iter()
+        .map(|field| {
```
theteachr, line 242: The body of this closure is big enough to have a name. Can we have an independent function?

trueleo: I refactored it. PTAL

theteachr (approved):

PR #579, opened 2023-12-19
Update dependencies

trueleo:
This PR has:


nitisht:
@trueleo fix the linter pls

trueleo:
> @trueleo fix the linter pls

There seems to be a behavior change in how infer_schema works. I will try to fix this.

PR #585, opened 2023-12-22
Add option for keeping local cache of recent data. 

trueleo:
### Description

This PR adds `local cache` / `hot tier` for Parseable.

This option can be enabled by setting following env vars
`P_CACHE_DIR` - Local Path for file cache
`P_CACHE_SIZE` - Size for cache in human readable size ( mb/mib/gib/gb )

When these flags are set, sync flow will move the parquet from staging into the specified cache directory instead of deleting it. Any LRU cached entry is deleted to satisfy the cache constraint upon insertion. LocalCacheManager is responsible for updating and persisting this data structure.

<hr>

This PR has:


server/src/option.rs
```diff
@@ -384,6 +398,25 @@ impl Server {
...
+                    .next_line_help(true),
+            )
+             .arg(
+                Arg::new(Self::CACHE_SIZE)
+                    .long(Self::CACHE_SIZE)
+                    .env("P_CACHE_SIZE")
+                    .value_name("mb")
+                    .default_value("1000")
+                    .value_parser(value_parser!(usize))
+                    .help("The local staging path is used as a temporary landing point for incoming events and local cache")
```
nitisht, line 418: Help text is wrong

server/src/option.rs
```diff
@@ -384,6 +398,25 @@ impl Server {
...
+                    .value_parser(validation::canonicalize_path)
+                    .help("The local staging path is used as a temporary landing point for incoming events and local cache")
+                    .next_line_help(true),
+            )
+             .arg(
+                Arg::new(Self::CACHE_SIZE)
+                    .long(Self::CACHE_SIZE)
+                    .env("P_CACHE_SIZE")
+                    .value_name("mb")
+                    .default_value("1000")
```
nitisht, line 416: 1000MB?

server/src/query/table_provider.rs
```diff
@@ -1,140 +0,0 @@
-/*
```
nitisht, line 1: Why deleted?

nitisht (changes requested):
Fix the commit message with more details on why we added this and how it works.

nitisht:
Missing items

- Prometheus stats on cache as we discussed.
- Cache info printed on the startup banner under storage.

nitisht:
What does this mapping in the .cache.json file mean?

```
[ 
"home/nitish/Development/parseablehq/parseable/data/teststream/date=2023-12-24/hour=04/minute=57/nitish-x1.data.parquet",
"/home/nitish/Development/parseablehq/parseable/cache/teststream/date=2023-12-24.hour=04.minute=57.nitish-x1.data.parquet"
]
```

nitisht:
If I restart the server and change the cache size via `P_CACHE_SIZE`, the `.cache.json` still has old value. We need to update the `.cache.json` so current specified size is honoured.

theteachr:
> What does this mapping in the .cache.json file mean?

@nitisht It's a mapping from store path to where that is stored in the local cache.

nitisht:
> What does this mapping in the .cache.json file mean?
> 
> ```
> [ 
> "home/nitish/Development/parseablehq/parseable/data/teststream/date=2023-12-24/hour=04/minute=57/nitish-x1.data.parquet",
> "/home/nitish/Development/parseablehq/parseable/cache/teststream/date=2023-12-24.hour=04.minute=57.nitish-x1.data.parquet"
> ]
> ```



> > What does this mapping in the .cache.json file mean?
> 
> @nitisht It's a mapping from store path to where the file is stored in the local cache.

That is clear, but

- Both files are local, still there is difference in path format - leading / vs no /
- Where do we use this mapping - that info should be added as comments and commit message

server/src/query/table_provider.rs
trueleo: This file is unused now

nitisht: Let's create a separate cleanup PR with details on why files are not used. this change is clearly unrelated to caching.

trueleo: >  This change is clearly unrelated to caching.

This file was being used to combine two execution plans ( remote and memory ), but now because of conditional addition of caching, I had to refactor related code in the schema provider file.

trueleo:
> Missing items
> 
>     * Prometheus stats on cache as we discussed.
> 
>     * Cache info printed on the startup banner under storage.

I will do this in a separate PR.

server/src/query/table_provider.rs
nitisht: Please fix commit message with details

nitisht:
> > Missing items
> > ```
> > * Prometheus stats on cache as we discussed.
> > 
> > * Cache info printed on the startup banner under storage.
> > ```
> 
> I will do this in a separate PR.

Let's finish in this PR

server/src/option.rs
```diff
@@ -606,4 +641,19 @@ pub mod validation {
...
+            s: &str,
+        ) -> Result<u64, human_size::ParsingError> {
+            SpecificSize::<T>::from_str(s).map(|x| x.to_bytes())
+        }
+
+        parse_and_map::<multiples::Mebibyte>(s)
+            .or(parse_and_map::<multiples::Megabyte>(s))
+            .or(parse_and_map::<multiples::Gigibyte>(s))
+            .or(parse_and_map::<multiples::Gigabyte>(s))
+            .map_err(|_| "Could not parse given size".to_string())
```
nitisht, line 657: let's add TB and Bytes as well.

server/src/storage/s3.rs
```diff
@@ -200,6 +200,12 @@ impl ObjectStorageProvider for S3Config {
         })
     }
 
+    fn get_store(&self) -> Arc<dyn ObjectStore> {
+        let s3 = self.get_default_builder().build().unwrap();
+        // limit objectstore to a concurrent request limit
+        Arc::new(LimitStore::new(s3, super::MAX_OBJECT_STORE_REQUESTS))
+    }
```
nitisht, line 207: Why is this needed - don't it being used?

server/src/localcache.rs
```diff
@@ -0,0 +1,189 @@
...
+    files: Cache<String, PathBuf>,
+}
+
+impl LocalCache {
+    fn new_with_size(capacity: u64) -> Self {
+        Self {
+            version: "v1".to_string(),
+            current_size: 0,
+            capacity,
+            files: Cache::new(100),
```
nitisht, line 47: Why 100 only here?

nitisht:
- Prometheus stats on cache as we discussed.
- Cache info printed on the startup banner under storage.
- Both files are local, still there is difference in path format - leading `/` vs no `/`
- Where do we use this mapping - that info should be added as comments and commit message
- If I restart the server and change the cache size via P_CACHE_SIZE, the .cache.json still has old value. We need to update the .cache.json so current specified size is honoured.
- What is the minimum allowed value for `P_CACHE_SIZE`? We need to fail if smaller values are set. IMO `1GiB` is a good number.

server/src/localcache.rs
trueleo: backing hashlru is capacity based. The capacity is arbitrary and is increased when needed.

nitisht: who increases? user or the backing cache?

nitisht (approved):

PR #588, opened 2023-12-24
Fix help messages for environment variables / flags

### LINKED ISSUE #586: bug: incorrect help message for several fields
Currently when displaying help, there are several incorrect descriptions for flags / environment vars. Also several env vars are not documented on the docs page https://www.parseable.io/docs/env-vars. 

This needs to be fixed.

```
➜  parseable git:(localcache) ./target/release/parseable local-store -h
Start Parseable with a drive as storage

Arguments:
  [filesystem path]  [env: P_FS_DIR=] [default: ./data]

Options:
      --tls-cert-path <PATH>
          The location of TLS Cert file [env: P_TLS_CERT_PATH=]
      --tls-key-path <PATH>
          The location of TLS Private Key file [env: P_TLS_KEY_PATH=]
      --address <ADDR:PORT>
          The address on which the http server will listen. [env: P_ADDR=] [default: 0.0.0.0:8000]
      --local-staging-path <DIR>
          The local staging path is used as a temporary landing point for incoming events and local cache [env: P_STAGING_DIR=] [default: ./staging]
      --cache-path <DIR>
          The local staging path is used as a temporary landing point for incoming events and local cache [env: P_CACHE_DIR=]
      --cache-size <mb>
          The local staging path is used as a temporary landing point for incoming events and local cache [env: P_CACHE_SIZE=] [default: 1000]
      --upload-interval <SECONDS>
          Interval in seconds after which un-committed data would be sent to the storage [env: P_STORAGE_UPLOAD_INTERVAL=] [default: 60]
      --username <STRING>
          Username for the basic authentication on the server [env: P_USERNAME=] [default: admin]
      --password <STRING>
          Password for the basic authentication on the server [env: P_PASSWORD=] [default: admin]
      --check-update <BOOL>
          Disable/Enable checking for updates [env: P_CHECK_UPDATE=] [default: true] [possible values: true, false]
      --send-analytics <BOOL>
          Disable/Enable sending anonymous user data [env: P_SEND_ANONYMOUS_USAGE_DATA=] [default: true] [possible values: true, false]
      --open-ai-key <STRING>
          Set OpenAI key to enable llm feature [env: P_OPENAI_API_KEY=]
      --oidc-client <STRING>
          Set client id for oidc provider [env: P_OIDC_CLIENT_ID=]
      --oidc-client-secret <STRING>
          Set client secret for oidc provider [env: P_OIDC_CLIENT_SECRET=]
      --oidc-issuer <URl>
          Set OIDC provider's host address. [env: P_OIDC_ISSUER=]
      --origin <URL>
          Set host global domain address [env: P_ORIGIN_URI=]
      --grpc-port <PORT>
          Set port for livetail arrow flight server [env: P_GRPC_PORT=] [default: 8001]
      --livetail-capacity <NUMBER>
          Set port for livetail arrow flight server [env: P_LIVETAIL_CAPACITY=] [default: 1000]
      --query-mempool-size <Gib>
          Set a fixed memory limit for query [env: P_QUERY_MEMORY_LIMIT=]
      --row-group-size <NUMBER>
          Number of rows in a row groups [env: P_PARQUET_ROW_GROUP_SIZE=] [default: 16384]
      --compression-algo <[UNCOMPRESSED, SNAPPY, GZIP, LZO, BROTLI, LZ4, ZSTD]>
          Parquet compression algorithm [env: P_PARQUET_COMPRESSION_ALGO=] [default: lz4] [possible values: uncompressed, snappy, gzip, lzo, brotli, lz4, zstd]
  -h, --help
          Print help
  -V, --version
          Print version
```

nitisht:
This PR also adds
* fix startup check for cache size and improve the error message
* gRPC port number in startup banner.
* Human readable input for size / capacity type environment variable

Fixes #586

This PR has:

### DIFF
--- a/server/src/option.rs
+++ b/server/src/option.rs
@@ -26,7 +26,7 @@ use std::sync::Arc;
 use url::Url;
 
 use crate::oidc::{self, OpenidConfig};
-use crate::storage::{FSConfig, ObjectStorageProvider, S3Config, LOCAL_SYNC_INTERVAL};
+use crate::storage::{FSConfig, ObjectStorageProvider, S3Config};
 use crate::utils::validate_path_is_writeable;
 
 pub const MIN_CACHE_SIZE_BYTES: u64 = 1000u64.pow(3); // 1 GiB
@@ -96,12 +96,6 @@ impl Config {
                 }
             }
             _ => unreachable!(),
-        }
-    }
-
-    pub fn validate(&self) {
-        if CONFIG.parseable.upload_interval < LOCAL_SYNC_INTERVAL {
-            panic!("object storage upload_interval (P_STORAGE_UPLOAD_INTERVAL) must be 60 seconds or more");
         }
     }
 
@@ -285,7 +279,7 @@ impl FromArgMatches for Server {
         self.livetail_channel_capacity = m
             .get_one::<usize>(Self::LIVETAIL_CAPACITY)
             .cloned()
-            .expect("default for livetail port");
+            .expect("default for livetail capacity");
         // converts Gib to bytes before assigning
         self.query_memory_pool_size = m
             .get_one::<u8>(Self::QUERY_MEM_POOL_SIZE)
@@ -380,7 +374,7 @@ impl Server {
                     .env("P_TLS_CERT_PATH")
                     .value_name("PATH")
                     .value_parser(validation::file_path)
-                    .help("The location of TLS Cert file"),
+                    .help("Local path on this device where certificate file is located. Required to enable TLS"),
             )
             .arg(
                 Arg::new(Self::TLS_KEY)
@@ -388,7 +382,7 @@ impl Server {
                     .env("P_TLS_KEY_PATH")
                     .value_name("PATH")
                     .value_parser(validation::file_path)
-                    .help("The location of TLS Private Key file"),
+                    .help("Local path on this device where private key file is located. Required to enable TLS"),
             )
             .arg(
                 Arg::new(Self::ADDRESS)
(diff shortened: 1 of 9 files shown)


server/src/option.rs
```diff
@@ -606,4 +605,22 @@ pub mod validation {
...
+                    "Invalid size provided. Please provide size in the format '1GiB', '3MiB', etc"
+                        .to_string(),
+                )
+            }
+        } else {
+            Err(
+                "Invalid size provided. Please provide size in the format '1GiB', '3MiB', etc"
+                    .to_string(),
+            )
+        }
```
theteachr, line 624:
```suggestion
    pub fn size(s: &str) -> Result<(), String> {
        // Unwrapping is Ok because the static expression is valid.
        let regex = regex::Regex::new(r"^(\d+)([TGMKP][iI]?[Bb]?)$").unwrap();

        if regex.is_match(s) {
            return Ok(());
        }

        return Err(
            "Invalid size provided. Please provide size in the format '1GiB', '3MiB', etc"
                .to_owned(),
        );
```

server/src/utils.rs
```diff
@@ -235,11 +236,65 @@ impl TimePeriod {
     }
 }
 
+pub fn parse_storage_size(size: Option<String>) -> Option<u64> {
```
theteachr, line 239: We'll use `human-size` crate introduced in https://github.com/parseablehq/parseable/pull/585.

theteachr (changes requested):

nitisht:
I'll clean this up once #585 is merged

PR #590, opened 2023-12-30
added a stream init time field

### LINKED ISSUE #587: Add first event date to logstream stats
It would me nice to be able to get the timestamp of the first event of the logstream in the stats in the api and console.

giripriyadarshan:
Fixes #587 .

### Description

adds a feature requested

as it is part of the stream's metadata, this was the only approach I thought of

adds a stream_init_time field to LogStreamMetadata struct. the field is initialized with current epoch time

<hr>

This PR has:
- [x] passes cargo build, and cargo test


giripriyadarshan:

server/src/handlers/http/logstream.rs
```diff
@@ -266,8 +266,16 @@ pub async fn get_stats(req: HttpRequest) -> Result<impl Responder, StreamError>
 
     let time = Utc::now();
 
+    let stream_init_time = &CONFIG
+        .storage()
+        .get_object_store()
+        .get_whole_json(&stream_name)
+        .await
+        .unwrap()["created-at"];
+
```
nitisht, line 275: We shouldn't have to get this from object store every time. There is an in-memory map maintained at Parseable server to avoid this exact same situation. Please use the map.

giripriyadarshan: resolved it in the latest commit .... now the created at data is loaded from the file either on load or once the file is created

nitisht:
@giripriyadarshan there are some more issues with the PR. We can discuss on Slack. Closing this one for now, in favor of #632

PR #601, opened 2024-01-04
feat: add first event timestamp in the stats api

### LINKED ISSUE #587: Add first event date to logstream stats

gurjotkaur20:
Fixes #587 

### Description
Added first event timestamp in stats api.

### DIFF
--- a/server/src/handlers/http/logstream.rs
+++ b/server/src/handlers/http/logstream.rs
@@ -21,7 +21,7 @@ use std::fs;
 use actix_web::http::StatusCode;
 use actix_web::{web, HttpRequest, Responder};
 use chrono::Utc;
-use serde_json::Value;
+use serde_json::{json, Value};
 
 use crate::alerts::Alerts;
 use crate::metadata::STREAM_INFO;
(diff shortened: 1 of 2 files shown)


nitisht (commented):
IMO it is too resource intensive to query the whole log stream everytime this API is called. A better approach would be to add a field `created_at` in the `LogStreamMetadata` struct.

gurjotkaur20:
> IMO it is too resource intensive to query the whole log stream everytime this API is called. A better approach would be to add a field `created_at` in the `LogStreamMetadata` struct.

I thought of this approach too but I just had one concern which timestamp would be there this field as the stream can be created separately without any log ingestion. 

- the timestamp when the stream is created? OR
- the timestamp when first log is ingested?

nitisht:
> > IMO it is too resource intensive to query the whole log stream everytime this API is called. A better approach would be to add a field `created_at` in the `LogStreamMetadata` struct.
> 
> I thought of this approach too but I just had one concern which timestamp would be there this field as the stream can be created separately without any log ingestion.
> 
> * the timestamp when the stream is created? OR
> * the timestamp when first log is ingested?

The issue #587 asks for the time stamp for the first event, But I think both the stream created time and the first event time - both are pretty useful.

PR #623, opened 2024-01-16
Proper Error Message if the Storage is in an invalid state

### LINKED ISSUE #615: error message: log proper error when S3 bucket has pre-existing data
It might get confusing if multiple applications are writing data to the same S3 bucket. Hence, Parseable server requires an empty and dedicated bucket to work on. 

Currently if the bucket has an existing prefix, error message looks like `Error: test/.stream.json not found`. This is confusing since what we mean here is that `test` prefix shouldn't be there at all.

Eshanatnight:
Fixes #615.

### Description

Return the proper Error Message if the Data Store is in an invalid state.

Invalid States
- `.parseable.json` does not exists but directories with `.stream.json` exist
- `.parseable.json` does not exists but directories with `.stream.json` do not exist
- `.parseable.json` does not exists and directories without `.stream.json` exists

Added a validation step to check if the Storage Location is in a valid state.

- Add new trait in `ObjectStorage`:  `async fn list_dirs_in_storage(&self) -> Result<Vec<String>, ObjectStorageError>`
- Impl list_dirs_in_storage for `localfs` and `s3`
- Add a validation method for checking if the Data Store is in a valid state.

<hr>

This PR has:
- [x ] been tested to ensure log ingestion and log query works.
- [ x] added comments explaining the "why" and the intent of the code wherever would not be obvious for an unfamiliar reader.

### DIFF
--- a/server/src/option.rs
+++ b/server/src/option.rs
@@ -111,20 +111,20 @@ impl Config {
         let has_parseable_json = obj_store.get_object(&rel_path).await.is_ok();
 
         let has_dirs = match obj_store.list_dirs_in_storage().await {
-            Ok(dirs) => dirs.is_empty(),
+            Ok(dirs) => !dirs.is_empty(),
             Err(_) => false,
         };
 
         let has_streams = obj_store.list_streams().await.is_ok();
 
-        if has_dirs || has_parseable_json && has_streams {
+        if !has_dirs || has_parseable_json && has_streams {
             Ok(())
         } else if has_parseable_json && !has_streams {
             Err(ObjectStorageError::Custom(
                 "Could not start the server because storage contains stale data from previous deployment, please choose an empty storage and restart the server. Join us on Parseable Slack to report this incident : launchpass.com/parseable"
                 .to_owned(),
             ))
-        } else if !has_parseable_json && !has_streams && !has_dirs {
+        } else if !has_parseable_json && !has_streams && has_dirs {
             Err(ObjectStorageError::Custom(
                 "Storage contains some stale data, Please provide an Empty Storage".to_owned(),
             ))
--- a/server/src/option.rs
+++ b/server/src/option.rs
@@ -126,7 +126,7 @@ impl Config {
             ))
         } else if !has_parseable_json && !has_streams && has_dirs {
             Err(ObjectStorageError::Custom(
-                "Could not start the server because storage contains some stale data, Please provide an Empty Storage.\nJoin us on Parseable Slack to report this incident : launchpass.com/parseable".to_owned(),
+                "Could not start the server because storage contains some stale data, please provide an empty storage and restart the server.\nJoin us on Parseable Slack to report this incident : launchpass.com/parseable".to_owned(),
             ))
         } else {
             Err(ObjectStorageError::Custom(
--- a/server/src/option.rs
+++ b/server/src/option.rs
@@ -27,7 +27,7 @@ use std::sync::Arc;
 use url::Url;
 
 use crate::oidc::{self, OpenidConfig};
-use crate::storage::{FSConfig, ObjectStorageProvider, S3Config};
+use crate::storage::{FSConfig, ObjectStorageError, ObjectStorageProvider, S3Config};
 use crate::utils::validate_path_is_writeable;
 
 pub const MIN_CACHE_SIZE_BYTES: u64 = 1000u64.pow(3); // 1 GiB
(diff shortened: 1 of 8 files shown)


server/src/option.rs
```diff
@@ -104,6 +104,37 @@ impl Config {
...
+            Err(_) => false,
+        };
+
+        let has_streams = obj_store.list_streams().await.is_ok();
+
+        if has_dirs || has_parseable_json && has_streams {
+            Ok(())
+        } else if has_parseable_json && !has_streams {
+            Err(ObjectStorageError::Custom(
+                "Parseable config found, but the Storage contains some stale data.".to_owned(),
```
nikhilsinhacloudsurfex, line 124: please correct the error to -
Could not start the server because storage contains stale data from previous deployment, please choose an empty storage and restart the server. Join us on Parseable Slack to report this incident : https://launchpass.com/parseable

server/src/option.rs
```diff
@@ -104,6 +104,37 @@ impl Config {
...
+            ))
+        } else if !has_parseable_json && !has_streams && !has_dirs {
+            Err(ObjectStorageError::Custom(
+                "Storage contains some stale data, Please provide an Empty Storage".to_owned(),
```
nikhilsinhacloudsurfex, line 128: Please provide the scenario when can this happen -!has_parseable_json && !has_streams && !has_dirs

server/src/option.rs
```diff
@@ -104,6 +104,37 @@ impl Config {
...
+            ))
+        } else {
+            Err(ObjectStorageError::Custom(
+                "Parseable config is missing, but streams are present in Storage.\nJoin us on Parseable Slack"
```
nikhilsinhacloudsurfex, line 132: Please provide the scenario when can this happen

server/src/option.rs
Eshanatnight: This error will be triggered if someone added some data to the storage where parseable stores data.
This case is the least likeliest to happen.

server/src/option.rs
Eshanatnight: The variable `has_dirs` does not properly convey the intention.

In this case `.parseable.json`, folders with `.stream.json` do not exists, but there is some data present in the storage.
This will happen if parseable is given a storage that is not empty.

server/src/option.rs
Eshanatnight: This is again a highly unlikely scenario to happen. 
This error will be triggered if `.parseable.json` was deleted from the storage manually.

server/src/option.rs
Eshanatnight: I am updating the logic of the if condition, so the intention is clearer.

server/src/option.rs
```diff
@@ -99,9 +98,36 @@ impl Config {
...
+        if !has_dirs || has_parseable_json && has_streams {
+            Ok(())
+        } else if has_parseable_json && !has_streams {
+            Err(ObjectStorageError::Custom(
+                "Could not start the server because storage contains stale data from previous deployment, please choose an empty storage and restart the server.\nJoin us on Parseable Slack to report this incident : launchpass.com/parseable"
+                .to_owned(),
+            ))
+        } else if !has_parseable_json && !has_streams && has_dirs {
+            Err(ObjectStorageError::Custom(
+                "Could not start the server because storage contains some stale data, please provide an empty storage and restart the server.\nJoin us on Parseable Slack to report this incident : launchpass.com/parseable".to_owned(),
```
nikhilsinhacloudsurfex, line 123: @Eshanatnight please update to below error

"Could not start the server because storage indicates stale data not related to parseable, please choose an empty storage and restart the server.\nJoin us on Parseable Slack to report this incident : launchpass.com/parseable"

server/src/option.rs
```diff
@@ -99,9 +98,36 @@ impl Config {
...
+            ))
+        } else {
+            Err(ObjectStorageError::Custom(
+                "Could not start the server because storage contains stale data from previous deployment.\nJoin us on Parseable Slack to report this incident : launchpass.com/parseable"
```
nikhilsinhacloudsurfex, line 127: @Eshanatnight please update to below error

"Could not start the server because storage indicates stale data from previous deployment, please choose an empty storage and restart the server.\nJoin us on Parseable Slack to report this incident : launchpass.com/parseable"

server/src/option.rs
```diff
@@ -99,9 +98,36 @@ impl Config {
...
+            Err(_) => false,
+        };
+
+        let has_streams = obj_store.list_streams().await.is_ok();
+
+        if !has_dirs || has_parseable_json && has_streams {
+            Ok(())
+        } else if has_parseable_json && !has_streams {
+            Err(ObjectStorageError::Custom(
+                "Could not start the server because storage contains stale data from previous deployment, please choose an empty storage and restart the server.\nJoin us on Parseable Slack to report this incident : launchpass.com/parseable"
```
nikhilsinhacloudsurfex, line 118: @Eshanatnight please update to below error message 

"Could not start the server because storage indicates stale data from previous deployment, please choose an empty storage and restart the server.\nJoin us on Parseable Slack to report this incident : launchpass.com/parseable"

nikhilsinhacloudsurfex (approved):
@Eshanatnight  please change the error messages as per the comments, rest all look good. I have tested the scenarios and all scenarios work as expected.
@nitisht once Eshan makes the error messages consistent, we are good to merge this PR as well.

nitisht:
Merging now, I'll fix the error message in next PR.

PR #624, opened 2024-01-16
Mutex Poison Error if a large amount of data is sent

Eshanatnight:
Fixes Writer Getting Error.

### DIFF
--- a/server/src/event/writer.rs
+++ b/server/src/event/writer.rs
@@ -38,7 +38,7 @@ pub static STREAM_WRITERS: Lazy<WriterTable> = Lazy::new(WriterTable::default);
 
 #[derive(Default)]
 pub struct Writer {
-    pub mem: MemWriter<16384>,
+    pub mem: MemWriter,
     pub disk: FileWriter,
 }
 
--- a/server/src/event/writer.rs
+++ b/server/src/event/writer.rs
@@ -78,10 +78,18 @@ impl WriterTable {
 
         match hashmap_guard.get(stream_name) {
             Some(stream_writer) => {
-                stream_writer
-                    .lock()
-                    .unwrap() // /yyyyyyyyyyy
-                    .push(stream_name, schema_key, record)?;
+                loop {
+                    match stream_writer.lock() {
+                        Ok(mut stream_writer) => {
+                            stream_writer.push(stream_name, schema_key, record)?;
+                            break;
+                        }
+                        Err(_ /*poisoned */) => {
+                            std::thread::sleep(Duration::from_millis(100));
+                            continue;
+                        }
+                    }
+                }
             }
             None => {
                 drop(hashmap_guard);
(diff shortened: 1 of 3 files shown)


server/src/event/writer.rs
nikhilsinhacloudsurfex, line 1: @Eshanatnight we discussed to make the change for line82 as well right?

Eshanatnight: I'm sorry it slipped my mind. I updated the file.

server/src/event/writer/mem_writer.rs
```diff
@@ -25,20 +25,22 @@ use itertools::Itertools;
 
 use crate::utils::arrow::adapt_batch;
 
+static BUF_SIZE: &str = "P_BUFFER_SIZE";
```
nikhilsinhacloudsurfex, line 28: @Eshanatnight did you put a default value of 16mb in case P_BUFFER_SIZE is not set

server/src/event/writer/mem_writer.rs
```diff
@@ -83,15 +85,20 @@ fn concat_records(schema: &Arc<Schema>, record: &[RecordBatch]) -> RecordBatch {
...
     pub inner: Vec<RecordBatch>,
     pub rows: usize,
 }
 
-impl<const N: usize> MutableBuffer<N> {
+impl MutableBuffer {
     fn push(&mut self, rb: RecordBatch) -> Option<Vec<RecordBatch>> {
-        if self.rows + rb.num_rows() >= N {
-            let left = N - self.rows;
+        let buf_size = std::env::var(BUF_SIZE)
```
nitisht, line 95: we also add a validation - i.e. use the standard route like other env vars (via clap)

server/src/event/writer/mem_writer.rs
nitisht, line 28: `P_BUFFER_SIZE` --> `P_RECORDS_PER_REQUEST`

Eshanatnight: Yes, that variable is just there so that Deepsource errors are not there

server/src/event/writer/mem_writer.rs
Eshanatnight: If we use get the value with clap we would need to add it to the `Server` struct. And will appear in cli options. Is that the behavior we want?

nitisht: I was thinking we should avoid putting this in the help etc, but the pattern of ad hoc env vars seems even more dangerous. So let's stick to Clap

Eshanatnight: Ok

Eshanatnight:
### Updated the Clap CLI 
CLI can now accept `records_per_request` and the default value is 16mb. 
Fallback to environment variable `P_RECORDS_PER_REQUEST`

PR #626, opened 2024-01-16
Fixes #616 feature: allow configurable duration for data push to S3

### LINKED ISSUE #616: feature: allow configurable duration for data push to S3
For some users, per minute might generate very small hence inefficient Parquet file, for some a minute worth data might already be too big a file.

Instead to allow better flexibility, the duration can be configurable. So user can choose how frequently they want data to be pushed to S3 in form of Parquet.

nikhilsinhacloudsurfex:
Fixes #616 

create parquet file by grouping all arrow files (in staging) for the duration provided in env variable P_STORAGE_UPLOAD_INTERVAL also check if arrow files vector is not empty,
then sort the arrow files and create key for parquet file from last file from sorted arrow files vector

### DIFF
--- a/server/src/storage/staging.rs
+++ b/server/src/storage/staging.rs
@@ -124,6 +124,7 @@ impl StorageDir {
         // hashmap <time, vec[paths]> but exclude where hot filename matches
         let mut grouped_arrow_file: HashMap<PathBuf, Vec<PathBuf>> = HashMap::new();
         let mut arrow_files = self.arrow_files();
+
         arrow_files.retain(|path| {
             !path
                 .file_name()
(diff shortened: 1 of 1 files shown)


server/src/storage/staging.rs
```diff
@@ -132,12 +133,17 @@ impl StorageDir {
...
                 .ends_with(&hot_filename)
         });
-        for arrow_file_path in arrow_files {
-            let key = Self::arrow_path_to_parquet(&arrow_file_path);
-            grouped_arrow_file
-                .entry(key)
-                .or_default()
-                .push(arrow_file_path);
+
+        //check if arrow files is not empty, fetch the parquet file path from last file from sorted arrow file list
```
sinhaashish, line 137: Try this out 

Don't use unwrap(). It panics in production.

```
if !arrow_files.is_empty() {
    arrow_files.sort();
    let key = Self::arrow_path_to_parquet(arrow_files.last().unwrap_or_default());

    let entry = grouped_arrow_file.entry(key.clone()).or_default();
    entry.extend(arrow_files);
}
```

nitisht (approved):

PR #635, opened 2024-01-25
Add functionality to read directly from arrow files

### LINKED ISSUE #631: query: staging data querying should happen directly from the disk
Currently the server maintains a small in-memory copy of events in the staging directory. So the query engine refers this in-memory copy in a query time range refers to staging data. This hampers future features that we want to build:

1. To build a configurable time range for staging duration. Current approach won't scale because the whole duration's data has to be kept in memory.
2. The in-memory staging is quite complicated, causing issue when large http payloads are sent to the server.

Staging query was added in the PR #448 

We need to instead query the staging data from disk itself.

Eshanatnight:
Fixes #631 .

### Description

Read Record Batches directly from arrow files and remove `MemWriter` Struct

#### Changes

- Complete Removal of `MemWriter`
- Update `Writer` and `WriterTable`structs

<hr>

This PR has:


nitisht:
we'll not take this as explained in #651

PR #640, opened 2024-01-29
Flatten OTEL logs to ingest on Parseable

### LINKED ISSUE #312: feature: support OpenTelemetry Protocol (OTLP)
I'm very interested in using parseable as it looks like a great answer to bloated/closed logging services, but I'm also keen to adopt OTLP for my projects in the hope that they could then be integrated seamlessly into different environments with different logging services (including, tragically, bloated/closed ones).

Is support for the OpenTelemetry Protocol (OTLP, [specification](https://opentelemetry.io/docs/reference/specification/protocol/otlp/)) on the roadmap at all? Would there be appetite to discuss it/merge a PR introducing it?

nikhilsinhacloudsurfex:
Fixes #312  

enhancement to add feature to flatten otel logs to be able to ingest on Parseable

<hr>

This PR has:


Eshanatnight:
I think it's better if we implement serde deserialize trait for the structs, or else it looks overwhelming

Eshanatnight:
> I think it's better if we implement serde deserialize trait for the structs, or else it looks overwhelming

I looked over on crates.io there are crates for open telemetry but, none provide the ability to serialize into the structs.

nikhilsinhacloudsurfex:
I also found some crates for otel but are vrl specific used by vector

Eshanatnight:
https://crates.io/crates/opentelemetry
There are more crates 
They define the data structures and functions that we won't need to maintain. I think it's worth checking it out.


And https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/logstransformprocessor

PR #647, opened 2024-02-01
update helm chart to correct domain

nitisht:


nitisht:
Skipping CI because none of the changes are related to Rust

PR #701, opened 2024-03-14
feat: Impl Query Server Changes

Eshanatnight:
### Description

1. Add: Schema API for Ingestion Server
2. RM: Redundant API Endpoints
3. feat: Last min data is Queryable
4. Add: Unit Tests
5. If the Query returns empty results, an empty vector will be returned rather than an error

<hr>

This PR has:

### DIFF
--- a/server/src/handlers/http.rs
+++ b/server/src/handlers/http.rs
@@ -17,6 +17,10 @@
  */
 
 use actix_cors::Cors;
+use arrow_schema::Schema;
+use serde_json::Value;
+
+use self::{modal::query_server::QueryServer, query::Query};
 
 pub(crate) mod about;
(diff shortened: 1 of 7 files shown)


Eshanatnight:
## Issues	

|No.| Description | Pointers|
|----|--------------|---------|
| 1.  | Staging Query Not working |  |
| 2. | Stream Sync on new  Start  | Stream Sync on startup |
| 3. |  if 2 Ingest Server are started, one is able to ingest, other is not. Erroring with  **`error - You don't have permission to access this resource. Please contact your administrator for assistance.`** | |

Eshanatnight:
## Changes

1. If the Query returns empty results, an empty vector will be returned rather than an error

PR #729, opened 2024-03-30
allow type casting in static schema

### LINKED ISSUE #639: Inquiry on Data Type Compatibility and Specification in Ingest API
Dear Team,

Firstly, I would like to express my respect for your hard work and dedication to our project. I am reaching out to discuss a matter I encountered with the `/api/v1/ingest` endpoint. It appears that during the initial stream setup, a column with a float64 datatype is created. However, when data like the integer '1' is ingested into this float64 column, a 400 error with the message "Invalid Request: Could not process this event due to mismatch in datatype" is returned. Interestingly, inputting the value '1.1' processes correctly.

To better understand this, I manually inserted data and verified the schema via the schema API, confirming that the 'in' field is indeed 'Float64'. This has led me to wonder if this is the intended behavior, especially since I am not fully versed in strict typing systems.

My primary inquiry is whether there is a method to explicitly specify types when ingesting event data or metadata. This information would greatly aid in aligning our data ingestion processes with the system’s requirements. Alternatively, considering and documenting support for forced typecasting in certain special cases would be very helpful.

Thank you for your time and assistance in this matter.

nikhilsinhacloudsurfex:
if field defined as float,
below values (examples) can be accepted -
100,-100.45, 100.23, "200.45"

Fixes #639 .

### Description

<hr>

This PR has:

### DIFF
--- a/server/src/event/format.rs
+++ b/server/src/event/format.rs
@@ -42,6 +42,7 @@ pub trait EventFormat: Sized {
         self,
         schema: HashMap<String, Arc<Field>>,
         time_partition: Option<String>,
+        static_schema_flag: Option<String>,
     ) -> Result<(Self::Data, EventSchema, bool, Tags, Metadata), AnyError>;
     fn decode(data: Self::Data, schema: Arc<Schema>) -> Result<RecordBatch, AnyError>;
     fn into_recordbatch(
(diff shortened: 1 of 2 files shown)


nitisht (approved):
LGTM!

aleimu:
==============================================================================
~~~ REQUEST ~~~
POST  /api/v1/ingest  HTTP/1.1
HOST   : 192.168.33.210:8002
HEADERS:
        Authorization: Basic YWRtaW46YWRtaW4=
        Content-Type: application/json
        User-Agent: go-resty/2.15.3 (https://github.com/go-resty/resty)
        X-P-Stream: charge_task_record_test_2
BODY   :
[
   {
      "id": 111,
      "action": "test",
      "user_id": 0,
      "soc": 1,
      "mileage": 123
   }
]

------------------------------------------------------------------------------
~~~ RESPONSE ~~~
STATUS       : 400 Bad Request
PROTO        : HTTP/1.1
RECEIVED AT  : 2024-11-13T10:36:40.978560864+08:00
TIME DURATION: 2.326225ms
HEADERS      :
        Content-Type: text/plain; charset=utf-8
        Date: Wed, 13 Nov 2024 02:36:40 GMT
        Vary: accept-encoding, Origin, Access-Control-Request-Method, Access-Control-Request-Headers
BODY         :
Invalid Request: Could not process this event due to mismatch in datatype
==============================================================================
sock file [/var/qlog_platform.sock] not found, uds log ignored

PR #732, opened 2024-04-02
fix analytics for the cluster

Eshanatnight:


Eshanatnight:
added active ingesters and offline ingesters to the metrics

nitisht (approved):

PR #738, opened 2024-04-03
cleaned up readme for better readability

nitisht:


nitisht:
Merging without CI, because text change only

PR #757, opened 2024-04-11
fixed liveness and readiness api to use base path

nikhilsinhaparseable:
Fixes #XXXX.

### Description

<hr>

This PR has:


nitisht:
cc @Eshanatnight did we change the path on purpose earlier?

nitisht:
> No the path was never under `api/v1` it was just `address/readiness` and `address/liveness`

It was under `api/v1/liveness` earlier: https://github.com/parseablehq/parseable/blob/b5acf0b827d5ff56dcceeb34a95c0adc3842c12b/server/src/handlers/http.rs#L323

Eshanatnight:
Yes I see that, I was a mistake on my end. I apologise

PR #761, opened 2024-04-16
update files in distributed mode to use hash

Eshanatnight:
~Needs to be rebased after #760 merge~

### DIFF
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -105,6 +105,7 @@ hashlru = { version = "0.11.0", features = ["serde"] }
 path-clean = "1.0.1"
 prost = "0.12.3"
 prometheus-parse = "0.2.5"
+sha2 = "0.10.8"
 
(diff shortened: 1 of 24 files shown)


server/Cargo.toml
```diff
@@ -105,6 +105,7 @@ hashlru = { version = "0.11.0", features = ["serde"] }
 path-clean = "1.0.1"
 prost = "0.12.3"
 prometheus-parse = "0.2.5"
+sha2 = "0.10.8"
```
nitisht, line 108: why not use [ulid](https://github.com/ulid/spec) that we use for deployment ids?

Eshanatnight: I am hashing the timestamp, so I needed sha2

nitisht:
We should add a check in querier for hash collision. I mean querier should
stop if it detects a collision- because this will lead to data loss
otherwise

PR #766, opened 2024-04-17
fix for delete stream API for distributed

### LINKED ISSUE #763: Fix: Delete Stream for Distributed
Querier -

deletes stream folder from S3

sends signal to ingesters' API

Ingester -

API to be updated - DELETE /logstream/{logstream}

On API call from querier - delete from STREAM_INFO

nikhilsinhaparseable:
change - querier deletes the stream folder from the storage 
then calls delete stream API for each ingester
ingester deletes the stream from its local map

Fixes #763 .


Eshanatnight (approved):
Looks good

PR #767, opened 2024-04-18
Fix: Cache fix distributed

### LINKED ISSUE #764: Fix: Caching for Distributed
Querier -

enable/disable caching for stream

sends signal to ingesters' API

Ingester -

API to be updated - PUT /logstream/{logstream}/cache

On API call from querier - enable/disable caching and write to its stream.json

nikhilsinhaparseable:
cache can be enabled from querier UI
querier calls the PUT /cache API to all ingesters 
at ingester, first checks if stream exists, 
if not found in local map, check in S3 and create stream 
then checks if caching env vars are set 
if yes, add cache_enabled flag to STREAM_INFO 
and update its stream.json in S3

Fixes #764

### DIFF
--- a/server/src/handlers/http/cluster/mod.rs
+++ b/server/src/handlers/http/cluster/mod.rs
@@ -48,6 +48,46 @@ use super::base_path_without_preceding_slash;
 
 use super::modal::IngestorMetadata;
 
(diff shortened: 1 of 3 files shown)


Eshanatnight (approved):
looks good

PR #775, opened 2024-04-22
fix: schema update where querying

Eshanatnight:
Fix schema update for complex queries


server/src/handlers/http/query.rs
```diff
@@ -63,27 +63,26 @@ pub struct Query {
...
+                Err(err) => {
+                    log::error!(
+                        "Failed to fetch schema for stream: {}, Error: {}",
+                        stream,
+                        err
+                    );
+                    continue;
+                }
+            }
         }
```
nitisht, line 86: Is this happening for each and every query? We should call this only when we get an error like field not found.

PR #786, opened 2024-04-29
Multiple Fixes for Distributed Deployments

nikhilsinhaparseable:
- to subtract the size of the arrow file removed from staging from staging size metrics
- to return error in case global cache is not set - /GET cache returns true/false but does not even check if global cache was enabled
change done is to return error if global cache is not set, if set return true/false
- query fix for migration from standalone to distributed deployments - query was not working for the manifest listed dates but was working when from date was given a day before the manifest listed date
change done is to add querier manifest.json also while checking for all manifests related to the time filter
- fix for edge case when ingestor was down/unreachable at the timeout of retention activity and ingestor's stream.json did not get updated now when ingestion happens from ingestor, local to storage sync fails with error - Manifest found in snapshot but not in object-storage change done is to create the manifest but not update the snapshot because snapshot already has the manifest entry

### DIFF
--- a/server/src/catalog.rs
+++ b/server/src/catalog.rs
@@ -34,9 +34,8 @@ use std::io::Error as IOError;
 pub mod column;
 pub mod manifest;
 pub mod snapshot;
-
+use crate::storage::ObjectStoreFormat;
 pub use manifest::create_from_parquet_file;
-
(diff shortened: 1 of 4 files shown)


Eshanatnight (commented):
Looks Good

PR #790, opened 2024-05-03
Allow Ingestion with Time Partition - Till 1 month older events

nikhilsinhaparseable:
update in the time partition logic
to check if time partition field value is not less than a month old
then partition based on time partition field
also, if static schema is provided
check if static schema has time partition field at the time of log creation
update the data type of time partition field to timestamp when persisting schema
this improves query performance of count(*)


Eshanatnight (commented):
looks good but quest seems to be failing

PR #795, opened 2024-05-08
enhancement to allow custom partitions in ingestion

nikhilsinhaparseable:
to allow - add an optional header - X-P-Custom-Partition in stream creation API with comma separated values 
the custom-partition will be stored in the stream.json file at the time of ingestion, 
below validations are put in place -
1. all events in the batch have the custom partitions
2. max of 3 partitions are allowed
3. custom partition values should not have a '.' in the log event
4. custom partition value should not be empty in the log event

Prefixes will be created in the alphabetical order of the custom partition fields

sort order in the parquet will be p_timestamp/time_partition and the custom partition fields same sort order will be reflected in the manifest file

### DIFF
--- a/server/src/event.rs
+++ b/server/src/event.rs
@@ -57,7 +57,7 @@ impl Event {
         }
 
         if !self.custom_partition_values.is_empty() {
-            let mut custom_partition_key = String::new();
+            let mut custom_partition_key = String::default();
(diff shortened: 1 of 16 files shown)


Eshanatnight (approved):
lgtm

PR #847, opened 2024-07-09
Fix: Server throws "User Does not have permission to access this" occasionally (arrow flight endpoint)

### LINKED ISSUE #822: Server throws "User Does not have permission to access this" occasionally (arrow flight endpoint).
[python-script.txt](https://github.com/user-attachments/files/16141168/script.txt)

requirements
`certifi==2024.2.2
charset-normalizer==3.3.2
grpcio==1.62.2
idna==3.7
numpy==1.26.4
ordered-set==4.1.0
pandas==2.2.2
py2wasm==2.6.2
pyarrow==15.0.2
python-dateutil==2.9.0.post0
pytz==2024.1
requests==2.31.0
six==1.16.0
tqdm==4.66.2
tzdata==2024.1
urllib3==2.2.1
zstandard==0.22.0
`

Steps to reproduce: 

1. Close all console instances in the browser
2. Restart the server.
3. Run the Python script (with a valid stream name and range)
4. Now open the console and login and go to any page
5. Run the same script again

Server throws: 
`pyarrow._flight.FlightUnauthorizedError: Flight returned unauthorized error, with message: User Does not have permission to access this. gRPC client debug context: UNKNOWN:Error received from peer ipv4:127.0.0.1:8002 {created_time:"2024-07-09T14:42:00.892379+05:30", grpc_status:7, grpc_message:"User Does not have permission to access this"}. Client context: IOError: Server never sent a data message. Detail: Internal`

dharanad:
## WIP

Fixes #822

### Description

When the parsable instance starts, it creates a session key for the admin user (of type `SessionKey::BasicAuth`) and adds it to the active session. This allows querying via Arrow Flight, as the session key is recognized as active.

However, when the admin user logs in, the admin session key (of type `SessionKey::BasicAuth`) is removed from the active session and new admin session of type `SessionKey::SessionId` is added. Consequently, any subsequent queries by clients using Arrow Flight fail the authentication step because the handler create a session key of type `SessionKey::BasicAuth` and tries to lookup in the map which fails for obvious reasons.

To address this, override permissions if the session key belongs to the user. This ensures that user sessions are maintained even if the admin logs in or logs out, preventing disruptions in client access via Arrow Flight.

### Other Approach 
* To maintain a mapping for `SessionKey::BasicAuth` to `SessionKey::SessionId`.  If map get operation fails for type `BasicAuth` we retrived the `SessionId` if exists and check if session id exists in active sessions.
* Implement a hash function which yeilds same hash value for BasicAuth and SessionId if they belong to same user.

<hr>

This PR has:

### DIFF
--- a/server/src/rbac.rs
+++ b/server/src/rbac.rs
@@ -105,7 +105,11 @@ impl Users {
     }
 
     pub fn remove_session(&self, session: &SessionKey) -> Option<String> {
-        mut_sessions().remove_session(session)
+        if !session.is_admin() {
+            mut_sessions().remove_session(session)
+        } else {
+            None
+        }
     }
 
     pub fn new_session(&self, user: &User, session: SessionKey) {
--- a/server/src/rbac.rs
+++ b/server/src/rbac.rs
@@ -105,11 +105,7 @@ impl Users {
     }
 
     pub fn remove_session(&self, session: &SessionKey) -> Option<String> {
-        if !session.is_admin() {
-            mut_sessions().remove_session(session)
-        } else {
-            None
-        }
+        mut_sessions().remove_session(session)
     }
 
     pub fn new_session(&self, user: &User, session: SessionKey) {
--- a/scripts/download.ps1
+++ b/scripts/download.ps1
@@ -7,7 +7,10 @@
 #  of them when querying for the architecture.
 #>
 [System.String[]]$SUPPORTED_ARCH = @("x86_64", "arm64", "amd64")
-[System.String]$DOWNLOAD_BASE_URL="cdn.parseable.com/"
+
+# Progress bar in powershell slows down downloads a lot, disable it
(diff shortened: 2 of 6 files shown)


dharanad:

server/src/rbac.rs
```diff
@@ -105,7 +105,11 @@ impl Users {
     }
 
     pub fn remove_session(&self, session: &SessionKey) -> Option<String> {
-        mut_sessions().remove_session(session)
+        if !session.is_admin() {
+            mut_sessions().remove_session(session)
+        } else {
+            None
```
dharanad, line 111: The only place where result of this function is in logout use case. So this might casuse issue. Thinking should we move this check to a layer above. Somehow we need to keep the admin session alive throughout the life time of the parseable service

nitisht:
@nikhilsinhaparseable PTAL

nikhilsinhaparseable (approved):
LGTM

PR #853, opened 2024-07-22
fix: update banner messaging to be more coherent

nitisht:


nitisht:
Merging because this is a text only change

PR #864, opened 2024-07-31
query optimization by remove table listing

nikhilsinhaparseable:
to support the  deployments older than 1 year, when Parseable did not have the concept of catalogues (manifest.json),
query module checks if the query time range has date without manifest in the storage if true (overlapping query), query result is served directly by listing all parquets in the time range
this slows down the performance and increases S3 cost

hence, the fix is to remove the check with assumption that all dates must have the catalogues and manifest must decide the execution plan for the query

still, to support the older deployments, we can write a separate tool to generate the manifest from the available parquets for a particular date to be run separately

### DIFF
--- a/server/src/query/stream_schema_provider.rs
+++ b/server/src/query/stream_schema_provider.rs
@@ -347,24 +347,6 @@ impl TableProvider for StandardTableProvider {
             merged_snapshot = object_store_format.snapshot;
         }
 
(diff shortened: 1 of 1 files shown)


nitisht:
Our plan here is to still allow querying older data - while cleaning up the complexity we have in the check right now.

nitisht:
Closing this for now, we have to take another stab at it.

PR #879, opened 2024-08-08
Enhancement: Add swagger API documentation

parmesant:
### Description

Adds API documentation using `utoipa` from within Rust code. Also generates an up-to-date openapi yaml spec as soon as the server starts up. 
To view the swagger page, go to `{endpoint}/swagger-ui/`

To build the docusaurus page follow these steps-
```bash
rm -rf build/ # Remove the old build directory (might be unnecessary)
yarn docusaurus clean-api-docs all #  remove pre-existing `mdx` files
yarn docusaurus gen-api-docs all # make new `mdx` files
yarn docusaurus build # build the server (or use whatever command you use to build production server)
npm run serve # Serve (or use production command)
```

Covered scopes-
- `logstream`
- `query`
- `ingestion`
- `user`
- `rbac`
- `liveness` (partial since can't add two variants of the same endpoint)
- `readiness` (partial since can't add two variants of the same endpoint)

#### Quirks
Request headers are specified as an attribute for `params` of a request and hence can't be optional (added a note in their description explaining their nature)

<hr>

This PR has:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3926,6 +3926,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "serde_yaml"
+version = "0.9.34+deprecated"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6a8b1a1a2ebf674015cc02edccce75287f1a0130d394307b36743c2f5d504b47"
+dependencies = [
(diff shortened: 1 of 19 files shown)


parmesant:

PR #883, opened 2024-08-09
feat: hot tier for pmeta

nikhilsinhaparseable:
allow enabling hot tier for the internal stream
with a size of 10 MiB
only if hot tier is configured globally

### DIFF
--- a/server/src/hottier.rs
+++ b/server/src/hottier.rs
@@ -25,6 +25,7 @@ use std::{
 
 use crate::{
     catalog::manifest::{File, Manifest},
+    handlers::http::cluster::INTERNAL_STREAM_NAME,
     metadata::{error::stream_info::MetadataError, STREAM_INFO},
     option::{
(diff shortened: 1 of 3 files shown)


server/src/hottier.rs
```diff
@@ -685,6 +686,22 @@ impl HotTierManager {
 
         Ok(None)
     }
+
+    pub async fn put_internal_stream_hot_tier(&self) -> Result<(), HotTierError> {
+        if CONFIG.parseable.hot_tier_storage_path.is_some()
+            && !self.check_stream_hot_tier_exists(INTERNAL_STREAM_NAME)
+        {
+            let mut stream_hot_tier = StreamHotTier {
+                size: "10485760".to_string(),
```
nitisht, line 695: Can you make it a constant and refer it here? Same for `available_size`

nitisht (changes requested):

nitisht (approved):

PR #885, opened 2024-08-10
enhancement: change sizes to bytes in hot tier

nikhilsinhaparseable:
in PUT /logstream/{logstream}/hottier
request body:

`
{
    "size": "10737418240"
}
`
the size is expected to be in bytes

GET /logstream/{logstream}/hottier
response body:
{
    "size": "1073741824",
    "used_size": "121798121",
    "available_size": "951943703",
    "oldest_date_time_entry": "2024-08-10 11:45:00"
}

size, used_size, available_size are sent in bytes

### DIFF
--- a/server/src/handlers/http/logstream.rs
+++ b/server/src/handlers/http/logstream.rs
@@ -1029,7 +1029,10 @@ pub async fn get_stream_hot_tier(req: HttpRequest) -> Result<impl Responder, Str
     }
 
     if let Some(hot_tier_manager) = HotTierManager::global() {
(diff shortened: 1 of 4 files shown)


balaji-jr:
Can we add the " bytes" as a suffix to all the sizes?

nitisht (changes requested):
As we discussed let's make it in sync with other API, i.e. 

- Add the " bytes" as a suffix to all the sizes e.g. `123324354 Bytes`
- Timezone info on the 'oldest_date_time_entry' e.g. `2011-08-12T20:17:46.384Z`

PR #888, opened 2024-08-14
fix: stats error for distributed system

nikhilsinhaparseable:
error occurs after migration from <1.2.0 to 1.4.0
stream.json has version v5
current code has logic to check for version
which is no longer required because of migration

update code to remove version validation


nitisht:
Changes added in #887. Closing this

PR #889, opened 2024-08-15
enhancement: accept other certificates

nikhilsinhaparseable:
env var P_OTHER_CERT_PATH accepts a directory path 
where user can keep all the certificates intended to be accepted by the server

### DIFF
--- a/server/src/cli.rs
+++ b/server/src/cli.rs
@@ -35,7 +35,7 @@ pub struct Cli {
     pub tls_key_path: Option<PathBuf>,
 
     /// The location of other certificates to accept
-    pub other_cert_path: Option<PathBuf>,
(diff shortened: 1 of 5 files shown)


server/src/cli.rs
```diff
@@ -175,6 +179,14 @@ impl Cli {
                     .value_parser(validation::file_path)
                     .help("Local path on this device where private key file is located. Required to enable TLS"),
             )
+            .arg(
+                Arg::new(Self::OTHER_CERT)
+                    .long(Self::OTHER_CERT)
+                    .env("P_OTHER_CERT_PATH")
```
nitisht, line 185: Can we change the variable to `P_TRUSTED_CA_CERTS_DIR`

nitisht:
Can you update the doc as well @nikhilsinhaparseable

PR #891, opened 2024-08-19
remove validation for hot tier with time partition

nikhilsinhaparseable:
now user can set hot tier for a stream with time partition removed the check


nitisht (commented):
Please confirm if changes needed to the doc here: https://www.parseable.com/docs/features/tiering#setup-hot-tier

Please send a PR if changes are needed to the docs.

PR #893, opened 2024-08-20
Fix LB timeout in `QueryServer`

vishalkrishnads:
### Description

The query server takes quite a while to process incoming requests. But, most load balancers timeout before the computation can be completed. 

As a temporary fix, this PR configures the query handler to send a one byte HTTP 202 response after waiting for `P_PROXY_TIMEOUT - 5` seconds if the computation hasn't finished. The client can then reach back out with the same query after a while. If the results are ready, the server responds, or it times out again. The result persists in the cache and can be reused for the same query later.

<hr>

This PR has:

### DIFF
--- a/server/src/handlers/http/query.rs
+++ b/server/src/handlers/http/query.rs
@@ -73,7 +73,7 @@ pub struct Query {
 
 pub async fn query_keep_alive(
     req: HttpRequest,
-    query_request: web::Json<Query>,
+    query_request: Query,
 ) -> Result<HttpResponse, Error> {
     let (mut tx, rx) = mpsc::channel::<Result<Bytes, Error>>(10);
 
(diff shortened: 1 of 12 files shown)


server/src/handlers/http/query.rs
```diff
@@ -64,81 +67,142 @@ pub struct Query {
...
+    );
+
+    // Check if the query is already being processed or completed
+    {
+        let query_map = query_map.lock().await;
+        if let Some(status) = query_map.get(&hash) {
+            match status {
+                QueryStatus::Processing => {
+                    // Wait for 55 seconds and check again
+                    sleep(Duration::from_secs(55)).await;
```
nitisht, line 97: this should be a env variable `P_PROXY_TIMEOUT` from user. Default to `60s`. Here we do `P_PROXY_TIMEOUT-5s`.

server/src/handlers/http/query.rs
```diff
@@ -64,81 +67,142 @@ pub struct Query {
...
+            }
+            Err(err) => {
+                log::error!("Error processing query: {:?}", err);
+                query_map.remove(&hash_clone);
+            }
+        }
+    });
+
+    // Wait for 55 seconds and respond with HTTP 202
+    sleep(Duration::from_secs(55)).await;
```
nitisht, line 146: same as above.

server/src/handlers/http/query.rs
vishalkrishnads: okay, will change.

PR #907, opened 2024-08-29
Handling SIGTERM for Distributed and Standalone Mode

AdheipSingh:
- Handling of SIGTERM. This is needed for readiness probe to work, so whenever kubelet sends a SIGTERM to pod, readiness handler response unavailable, this way service will send no traffic to this pod. 

- This PR also makes sure actix drains HTTP connection before exiting. 

```
how to stop traffic. The current issue is service is still sending traffic to terminating pods. We added pre-stop hooks to sleep for kube proxy and core dns cache to clear, which worked at times, but failed too. So now, we want to make sure connections are drained.
Handling SIGTERM
1. Kubelet sends SIGTERM to pod when we issue delete pod
2. Actix hijacks the call ( thread is running ) and tweaks a global flag SIGNAL_RECEIVED to true.
3.Readiness Probe handler reads the flag each time, it is called.
4. on pod readiness probe is set, kubelet will call probe every 5 seconds.
5. Readiness Probe will return unavailable, and pod will go 0/1 unavailable.
6. Service at this point will route to other available pods.
7. After 120 seconds pod will terminate, during that period ingestor will s3 sync ( TODO: once this is working on k8s i will add in SIGTERM handling itself )
```

### Description

<hr>

This PR has:


AdheipSingh:
- I need to build the image from Dockerfile and test in cluster once again. I have been testing using Dockerfile.debug image.
This will also need changes in existing helm charts. To add readiness probe.
Please suggest if you want to add in same PR or new one. 
cc @nitisht @nikhilsinhaparseable

Dockerfile
```diff
@@ -34,4 +34,4 @@ WORKDIR /parseable
 COPY --from=builder /bin/sh /bin/sh
```
AdheipSingh, line 34: @nitisht any specific reason for adding shell. Is it because image is distroless, i have been using Dockerfile.debug, so just want to confirm ?

server/src/handlers/http/health_check.rs
```diff
@@ -1,34 +1,74 @@
...
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- * GNU Affero General Public License for more details.
- *
- * You should have received a copy of the GNU Affero General Public License
- * along with this program.  If not, see <http://www.gnu.org/licenses/>.
- *
- */
-
+use crate::option::CONFIG;
```
nitisht, line 1: pls add back the license header.

server/src/handlers/http/health_check.rs
```diff
@@ -1,34 +1,74 @@
...
+            }
+
+            log::info!("Local and S3 Sync done, handler SIGTERM completed.");
+        }
+        None => {
+            println!("Signal handler received None, indicating an error or end of stream");
+        }
+    }
+
+    eprintln!("Signal handler task completed");
```
nitisht, line 58: please remove all the `println` and `eprintln`.

nitisht:
> Please suggest if you want to add in same PR or new one.

Let's add in the same PR

server/src/handlers/http/modal/query_server.rs
```diff
@@ -74,16 +76,47 @@ impl ParseableServer for QueryServer {
...
+        // Spawn the signal handler task
+        tokio::spawn(async move {
+            health_check::handle_signals(shutdown_signal).await;
+            println!("Received shutdown signal, notifying server to shut down...");
+        });
+
+        // Create the HTTP server
+        let http_server = HttpServer::new(create_app_fn)
+            .workers(num_cpus::get())
+            .shutdown_timeout(60);
```
AdheipSingh, line 95: Hardcoded to 60 seconds. @nikhilsinhaparseable @nitisht please suggest if needed more.

Dockerfile
nitisht: Yes, because distroless doesn't have any other shell.

server/src/handlers/http/modal/query_server.rs
nitisht: we should make it 120 I think

Dockerfile
AdheipSingh: Hmm, but do we need a shell ? I mean wasn't shell for ${hostname}, container and app PID should be same. 
I'lll check once if needed. If you see CMD will trigger the command itself.

nitisht: Sure, we can remove in that case

server/src/handlers/http/health_check.rs
AdheipSingh: done

server/src/handlers/http/modal/query_server.rs
AdheipSingh: done

server/src/handlers/http/health_check.rs
AdheipSingh: done, added log::info.

AdheipSingh:
> > Please suggest if you want to add in same PR or new one.
> 
> Let's add in the same PR

Done

AdheipSingh:
TODO: to be tested with Dockerfile changes in cluster.

AdheipSingh:
Dockefile changes are testing. No shell copy needed.

nitisht (approved):

PR #910, opened 2024-09-01
hot tier details in banner and about api response

nikhilsinhaparseable:


nitisht:
Can you paste the banner here?

PR #941, opened 2024-09-25
Support SSE-C for S3 object storage

### LINKED ISSUE #919: Support encryption on object storage
Is there a technical reason for the omission of encryption on object storage (S3) ?

This could be done even without introducing any extra code to `parseable` if you exposed support for S3-side features such as [SSE-C](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ServerSideEncryptionCustomerKeys.html).

MihirLuthra:
Fixes #919.

This PR introduces new args/envs allowing users to use [SSE-C](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ServerSideEncryptionCustomerKeys.html) for encrypting the objects in `S3`.

To achieve that, an arg `--object-sse` and env `P_OBJECT_SSE` have been exposed. Example usage:
```
parseable s3-store --object-sse SSE-C:AES256:$BASE64_ENCRYPTION_KEY
```
Fortunately, `object_store` crate had already added a way to define `SSE-C` key from `AmazonS3Builder`: apache/arrow-rs#6230. But they have not published a new version to crates.io containing this change.

Although, even if they did, it won't be usable yet. Apparently, the version of datafusion being used in `parseable` is not compatible with anything above `0.10.2` yet. So, for now, I have created a fork of `arrow-rs` with a branch checked out of `object_store_0.10.2` tag. On top of that, I have cherry picked commits from:
- https://github.com/apache/arrow-rs/pull/6230
- https://github.com/apache/arrow-rs/pull/6260
Can be seen here: https://github.com/MihirLuthra/arrow-rs/tree/mihir/I-919-0.10.2-with-sse-c. This version of arrow-rs has been used in `parseable` by `[patch.crates-io]`.

In the changes here, I have used my own fork but if this way is acceptable, a fork would be needed in `parseablehq` organization instead.

Other ways could be:

- Fix datafusion to work with `0.11.0` and ask `arrow-rs` maintainers to publish new version on crates.io. I think this would be a new issue in itself.
- Ask arrow-rs maintainers if they are okay with backporting `SSE-C` change to 0.10 and introduce a new 0.10.3.

I have tested this with my AWS Account. If parseable is started with `SSE-C` configured, then an attempt to download object without the key looks like this:
```
❯ aws s3api get-object  --bucket parseable-bucket-2 --key demo/date=2024-09-24/hour=23/minute=46/mihir.data.czlkrB4wsZ6vf2X.parquet /dev/stdout | more

An error occurred (InvalidRequest) when calling the GetObject operation: The object was stored using a form of Server Side Encryption. The correct parameters must be provided to retrieve the object.
```
But following works if correct key and md5 hash given:

```
❯ aws s3api get-object  --bucket parseable-bucket-2 --key demo/date=2024-09-24/hour=23/minute=46/mihir.data.czlkrB4wsZ6vf2X.parquet /dev/stdout --sse-customer-algorithm AES256 --sse-customer-key $ENCRYPTION_KEY --sse-customer-key-md5 $ENCRYPTION_KEY_MD5 | more
```

For my testing, keys were generated as:

```bash
#!/bin/bash

# Generate a 256-bit (32-byte) AES key and base64 encode it
ENCRYPTION_KEY=$(openssl rand -base64 32)

# Compute the MD5 hash of the encryption key and base64 encode it
ENCRYPTION_KEY_MD5=$(echo -n $ENCRYPTION_KEY | base64 -d | openssl md5 -binary | base64)

echo "Encryption Key: $ENCRYPTION_KEY"
echo "Key MD5: $ENCRYPTION_KEY_MD5"
```

<hr>

This PR has:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2962,8 +2962,7 @@ dependencies = [
 [[package]]
 name = "object_store"
 version = "0.11.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "25a0c4b3a0e31f8b66f71ad8064521efa773910196e2cde791436f13409f3b45"
+source = "git+https://github.com/parseablehq/arrow-rs.git?rev=23b6ff9f432e8e29c08d47a315ba0b7cb8758225#23b6ff9f432e8e29c08d47a315ba0b7cb8758225"
 dependencies = [
  "async-trait",
  "base64 0.22.0",
@@ -3447,8 +3446,8 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0c1318b19085f08681016926435853bbf7858f9c082d0999b80550ff5d9abe15"
 dependencies = [
  "bytes",
- "heck 0.5.0",
- "itertools 0.13.0",
+ "heck 0.4.1",
+ "itertools 0.12.1",
  "log",
  "multimap",
  "once_cell",
@@ -3468,7 +3467,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e9552f850d5f0964a4e4d0bf306459ac29323ddfbae05e35a7c0d35cb0803cc5"
 dependencies = [
  "anyhow",
- "itertools 0.13.0",
+ "itertools 0.12.1",
  "proc-macro2",
  "quote",
  "syn 2.0.79",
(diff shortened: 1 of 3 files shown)


nikhilsinhaparseable:
@MihirLuthra parseable is now using object_store version 0.11.0, can you please update this PR to use the same.
Also, can you use https://github.com/parseablehq/arrow-rs instead of your fork for the ssec changes (this is only temporary, and will be used till the object_store release does not come with the changes).
Please update your PR with above mentoned comments.

nikhilsinhaparseable:
@MihirLuthra you can do that, no need to use fork then.

nikhilsinhaparseable:
@MihirLuthra I have renamed the env var and related variables in the change.
Otherwise, the change looks good and ready to merge (once all checks are passed)

Thanks for the enhancement!

nikhilsinhaparseable (approved):
looks good to merge

PR #964, opened 2024-10-20
Support Azure Blob Storage on AKS via access key

AdheipSingh:
Fixes #XXXX.

### Description

<hr>

This PR has:

### DIFF
--- a/helm/templates/querier-statefulset.yaml
+++ b/helm/templates/querier-statefulset.yaml
@@ -45,8 +45,13 @@ spec:
         imagePullPolicy: {{ .Values.parseable.image.pullPolicy }}
         args: 
         - /usr/bin/parseable
-        - s3-store
+        - {{ .Values.parseable.objectStore }}
         env:
+          - name: HOSTNAME
+            valueFrom:
+              fieldRef:
+                apiVersion: v1
+                fieldPath: metadata.name
           {{- range $key, $value :=  .Values.parseable.env }}
           - name: {{ $key }}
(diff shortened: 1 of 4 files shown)


helm/values.yaml
```diff
@@ -6,6 +6,9 @@ parseable:
   ## Set to true if you want to deploy Parseable in local mode (store logs
   ## on local mount point instead of S3 bucket)
   local: false
+  ## object store can be s3, blob
+  ## local needs to be false if set to object store.
+  objectStore: s3-store
```
nitisht, line 11: Shall we instead change it to 

```
store: local / s3-store / blob-store
```

AdheipSingh:
tested on azure.

helm/templates/querier-statefulset.yaml
```diff
@@ -60,11 +65,6 @@ spec:
               secretKeyRef:
                 name: {{ $secret.name }}
                 key: {{ $key }}
-          - name: HOSTNAME
-            valueFrom:
-              fieldRef:
-                apiVersion: v1
-                fieldPath: metadata.name
```
nitisht, line 67: This is needed, why remove this?

AdheipSingh: https://github.com/AdheipSingh/parseable/blob/aks-helm/helm/templates/querier-statefulset.yaml#L50
This is present just refactored it

nitisht: oh right, my bad.

PR #970, opened 2024-10-22
perf: do FAT LTO and singular `codegen-unit` on release

### LINKED ISSUE #224: PGO + Bolt
parseable right now does not support building with more advanced optimization techniques like [PGO](https://en.wikipedia.org/wiki/Profile-guided_optimization) and [BOLT](https://github.com/llvm/llvm-project/tree/main/bolt). This tooling has an increasing adoption in the community as a tool to additionally optimize programs. With this tooling, there is a huge chance to gain even more performance "for free".

Here I suggest considering an option at least to play with LTO ("fat" version) + PGO + Bolt pipeline (or any combination of them) and test, does it give a performance to the project or not. If yes, would be awesome to have prebuilt binaries with more advanced optimization from the scratch. Also, for the users will be helpful to have the ability to tweak manually their own binaries to their own workloads with the integrated into the build scripts functionality.

Since the project is quite small I do not expect (at least yet) significant improvement from the PGO right now. But at least a good option to try :)

Also, there are some caveats to consider like:
- Significantly increased build times
- BOLT could be still unstable (or even broken) on some architectures

Links:
- Rust experience with LTO + PGO + BOLT: https://kobzol.github.io/rust/rustc/2022/10/27/speeding-rustc-without-changing-its-code.html

de-sh:
Fixes #224 .

### Description

Optimize release builds by building them in a special profile `release-lto` that:
1. Enables `FAT LTO`
2. Uses `codegen-unit = 1`
This is intended to improve the quality of parseable binary compiled at release, in terms of performance and size.

<hr>

This PR has:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -127,3 +127,8 @@ assets-sha1 = "9d5a45f204d709a2dd96f6a5e0b21b3834ee0e36"
 
 [features]
 debug = []
+
+[profile.release-lto]
+inherits = "release"
+lto = "fat"
+codegen-units = 1
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -37,11 +37,11 @@ jobs:
         with:
           use-cross: true
           command: build
-          args: --release --target ${{matrix.target}}
+          args: --profile release-lto --target ${{matrix.target}}
 
       - name: Rename binary
         run: |
-          mv target/${{ matrix.target }}/release/parseable Parseable_OSS_${{ matrix.target }}
+          mv target/${{ matrix.target }}/release-lto/parseable Parseable_OSS_${{ matrix.target }}
 
       - name: Generate artifact attestation
         uses: actions/attest-build-provenance@v1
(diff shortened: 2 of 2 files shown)


de-sh:

Cargo.toml
```diff
@@ -8,3 +8,7 @@ resolver = "2"
 # - https://github.com/apache/arrow-rs/pull/6260
 # But a new version hasn't been published to crates.io for this yet. So, we are using this patch temporarily.
 object_store = { git = "https://github.com/apache/arrow-rs.git", rev = "23b6ff9f432e8e29c08d47a315ba0b7cb8758225" }
+
+[profile.release]
+lto = "thin"
+codegen-units = 1
```
nitisht, line 14: I think it is worth trying with FAT as well and see if we get any perf advantages. cc @nikhilsinhaparseable

Cargo.toml
```diff
@@ -8,3 +8,7 @@ resolver = "2"
 # - https://github.com/apache/arrow-rs/pull/6260
 # But a new version hasn't been published to crates.io for this yet. So, we are using this patch temporarily.
 object_store = { git = "https://github.com/apache/arrow-rs.git", rev = "23b6ff9f432e8e29c08d47a315ba0b7cb8758225" }
+
+[profile.release]
+lto = "thin"
```
de-sh, line 13:
```suggestion
```

Just realized that `"thin"` is default 🃏

de-sh:
Decision on this front, we will move these changes to a specific profile `release-lto` that will be used during compilation for binary generation on GitHub release.

nitisht (approved):

PR #971, opened 2024-10-22
chore: update `object_store`

de-sh:
Ref #941

### Description

Removes patch, updates to latest version of crate `object_store`.

<hr>

This PR has:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2961,8 +2961,9 @@ dependencies = [
 
 [[package]]
 name = "object_store"
-version = "0.11.0"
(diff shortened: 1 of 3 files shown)


nitisht (commented):
Shouldn't we update object_store crate also here to https://github.com/apache/arrow-rs/releases/tag/object_store_0.11.1

Cargo.lock
```diff
@@ -2961,8 +2961,9 @@ dependencies = [
 
 [[package]]
 name = "object_store"
-version = "0.11.0"
-source = "git+https://github.com/apache/arrow-rs.git?rev=23b6ff9f432e8e29c08d47a315ba0b7cb8758225#23b6ff9f432e8e29c08d47a315ba0b7cb8758225"
+version = "0.11.1"
```
de-sh, line 2964: @nitisht this does exactly that

de-sh: But I understand what you mean, have added a further commit

PR #974, opened 2024-10-25
refactor: use middlewares to reject requests when shutting down

de-sh:
Fixes #XXXX.

### Description

The idea behind this PR is to not have multiple copies of the shutdown handler code and as a **BREAKING CHANGE** it **WILL** result in all requests to ingest/query/server to be rejected with HTTP 503 status when `SIGNAL_RECEIVED` is set to true.

<hr>

This PR has:

### DIFF
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -25,7 +25,7 @@ tower-http = { version = "0.6.1", features = ["cors"] }
 
 ### actix dependencies
 actix-web-httpauth = "0.8"
-actix-web = { version = "4.5.1", features = ["rustls-0_22"] }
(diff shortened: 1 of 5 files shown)


server/Cargo.toml
```diff
@@ -25,7 +25,7 @@ tower-http = { version = "0.6.1", features = ["cors"] }
 
 ### actix dependencies
 actix-web-httpauth = "0.8"
-actix-web = { version = "4.5.1", features = ["rustls-0_22"] }
+actix-web = { version = "4.9.0", features = ["rustls-0_22"] }
```
de-sh, line 28: previous `from_fn` was part of an experimental crate, now it is in main

PR #975, opened 2024-10-25
detect datetime column on ingestion

balaji-jr:
For streams with dynamic schema, set datatype as Timestamp for new string fields if it contains valid datetime value.
And minor tweaks for detecting schema.


nikhilsinhaparseable (approved):
good to merge

PR #981, opened 2024-11-02
refactor: clean up parts of the codebase

de-sh:
Fixes #XXXX.

### Summary

- De-serialize `Mode`, don't compare strings.
- `to_string` runs `to_vec`, so just do that.
- Custom `SSEError` type for readability.
- Fewer hard-coded serde renames.
- fix: there are a 1000milis every sec.
- De-serialize `Compression` to not hard-code matching.
- Space used/available in hottier is not optional.
- Only construct the `ObjectStorage` client once.
- Reorganize as methods to save on cost of copy.
- Mutate only the `file_path` on windows. 
- `DiskUtil` replaces the tuple `(total_disk_space, available_disk_space, used_disk_space)`.
- Save on cost of cloning within `EventFormat`

<hr>

This PR has:


server/src/handlers/http/modal/ingest_server.rs
```diff
@@ -480,7 +373,12 @@ impl IngestServer {
 
         tokio::spawn(airplane::server());
 
-        let app = self.start(prometheus, CONFIG.parseable.openid.clone());
+        // set the ingestor metadata
+        self.set_ingestor_metadata().await?;
+
+        // TODO: figure out why we don't need this
```
de-sh, line 379: Why is it that only ingest doesn't care about OpenId?

de-sh: Also why is it that it is the only server that sets it's metadata?

nitisht: Ingest is almost always done via a software like an agent, or Kafka or an application and so on. OAuth based authentication will be a challenge for the workflow

server/src/handlers/http/modal/mod.rs
```diff
@@ -26,31 +26,145 @@ pub mod utils;
...
+        // Spawn the signal handler task
+        let signal_task = tokio::spawn(async move {
+            health_check::handle_signals(shutdown_signal).await;
+            println!("Received shutdown signal, notifying server to shut down...");
+        });
+
+        // Create the HTTP server
+        let http_server = HttpServer::new(create_app_fn)
+            .workers(num_cpus::get())
+            .shutdown_timeout(60);
```
de-sh, line 117: what should be shutdown timeout?

server/src/handlers/http/modal/ingest_server.rs
```diff
@@ -38,138 +44,44 @@ use crate::storage::ObjectStorageError;
...
-        // Spawn the signal handler task
-        let signal_task = tokio::spawn(async move {
-            health_check::handle_signals(shutdown_signal).await;
-            log::info!("Received shutdown signal, notifying server to shut down...");
-        });
-
-        // Create the HTTP server
-        let http_server = HttpServer::new(create_app_fn)
-            .workers(num_cpus::get())
-            .shutdown_timeout(60);
```
de-sh, line 123: 60 here

server/src/handlers/http/modal/query_server.rs
```diff
@@ -17,136 +17,55 @@
...
-        // Spawn the signal handler task
-        let signal_task = tokio::spawn(async move {
-            health_check::handle_signals(shutdown_signal).await;
-            log::info!("Received shutdown signal, notifying server to shut down...");
-        });
-
-        // Create the HTTP server
-        let http_server = HttpServer::new(create_app_fn)
-            .workers(num_cpus::get())
-            .shutdown_timeout(120);
```
de-sh, line 100: 120 here

server/src/handlers/http/modal/server.rs
```diff
@@ -62,133 +56,13 @@ use crate::{
...
-        // Spawn the signal handler task
-        let signal_task = tokio::spawn(async move {
-            health_check::handle_signals(shutdown_signal).await;
-            log::info!("Received shutdown signal, notifying server to shut down...");
-        });
-
-        // Create the HTTP server
-        let http_server = HttpServer::new(create_app_fn)
-            .workers(num_cpus::get())
-            .shutdown_timeout(60);
```
de-sh, line 120: What'd be the right value?

de-sh:

server/src/main.rs
parmesant, line 1: What difference does using a Boxed pointer makes here? Is it for legibility?

de-sh:

src/hottier.rs
```diff
@@ -51,15 +51,13 @@ pub const MIN_STREAM_HOT_TIER_SIZE_BYTES: u64 = 10737418240; // 10 GiB
...
 #[derive(Debug, serde::Deserialize, serde::Serialize)]
 pub struct StreamHotTier {
     pub version: Option<String>,
-    #[serde(rename = "size")]
     pub size: String,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub used_size: Option<String>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub available_size: Option<String>,
+    pub used_size: String,
```
de-sh, line 59: @nikhilsinhaparseable why was this optional previously, we keep unwrapping it everywhere, mostly before setting it to another value, is this redundant?

nitisht (approved):
LGTM!

PR #994, opened 2024-11-19
feat: audit logs

### LINKED ISSUE #765: feat: audit logging for Parseable
Audit logging is a core requirement for enterprise software. We need to add a feature to ensure a Parseable server (in any mode - local, s3, standalone, distributed) sends out audit logs to a target specified via env vars. 

Env variables to add:

- `P_AUDIT_LOG_TARGET` --> `String`, e.g. `https://target.domain.name`
- `P_AUDIT_LOG_TARGET_USERNAME` --> `String`, e.g. `user`
- `P_AUDIT_LOG_TARGET_PASSWORD` --> `String`, e.g. `password`
- `P_AUDIT_LOG_TARGET_TLS_VERIFY` --> `Boolean`, e.g. `false`
- `P_AUDIT_LOG_TARGET_HEADERS`  --> Command separated key value pair (`String:String,String:String`), e.g. `content-type:application/json, x-p-stream:audit`

Also, we need to decide a schema of each audit log event. It should include `version`, `eventType` and other relevant details.

Anirudhxx:
Opening this PR so it makes it easier to discuss adding audit logs to parseable.

Fixes #765.

What kind of events do we want to capture?
- Right now I am only tracking the Action events 
- Since these are accessible in the auth middleware I have sent the actual audit log post request in the auth middleware itself
- This most likely add a performance penalty to our server and should be moved elsewhere (Looking for some suggestions)

Adding any other details to the audit log
- Right now I am only capturing details like <action, ip address, request type>

This PR has:

### DIFF
--- a/server/src/cli.rs
+++ b/server/src/cli.rs
@@ -16,8 +16,8 @@
  *
  */
 
-use clap::{value_parser, Arg, ArgGroup, Command, FromArgMatches};
-use std::path::PathBuf;
+use clap::{builder::ValueParser, value_parser, Arg, ArgGroup, Command, FromArgMatches};
+use std::{collections::HashMap, path::PathBuf};
 
 use url::Url;
 
@@ -119,8 +119,28 @@ pub struct Cli {
     pub trino_auth: Option<String>,
     pub trino_schema: Option<String>,
     pub trino_catalog: Option<String>,
+
+    // audit log vars
+    pub audit_log_target: Option<String>,
+    pub audit_log_target_username: Option<String>,
+    pub audit_log_target_password: Option<String>,
+    pub audit_log_target_tls_verify: bool,
+    pub audit_log_target_headers: HashMap<String, String>,
 }
 
+fn parse_header(header: &str) -> Result<HashMap<String, String>, String> {
+    let mut map = HashMap::new();
+
+    for pair in header.split(',') {
+        if let Some((key, value)) = pair.split_once(':') {
+            map.insert(key.trim().to_string(), value.trim().to_string());
+        } else {
+            return Err(format!("Invalid header format: {}", pair));
+        }
+    }
+
+    Ok(map)
+}
 impl Cli {
     // identifiers for arguments
     pub const TLS_CERT: &'static str = "tls-cert-path";
@@ -163,6 +183,13 @@ impl Cli {
     pub const TRINO_USER_NAME: &'static str = "p-trino-user-name";
     pub const TRINO_AUTHORIZATION: &'static str = "p-trino-authorization";
     pub const TRINO_SCHEMA: &'static str = "p-trino-schema";
+
+    // audit log env vars
+    pub const P_AUDIT_LOG_TARGET: &'static str = "p-audit-log-target";
+    pub const P_AUDIT_LOG_TARGET_USERNAME: &'static str = "p-audit-log-target-username";
+    pub const P_AUDIT_LOG_TARGET_PASSWORD: &'static str = "p-audit-log-target-password";
+    pub const P_AUDIT_LOG_TARGET_TLS_VERIFY: &'static str = "p-audit-log-target-tls-verify";
+    pub const P_AUDIT_LOG_TARGET_HEADERS: &'static str = "p-audit-log-target-headers";
 
     pub fn local_stream_data_path(&self, stream_name: &str) -> PathBuf {
         self.local_staging_path.join(stream_name)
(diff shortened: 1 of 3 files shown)


server/src/cli.rs
```diff
@@ -119,8 +119,28 @@ pub struct Cli {
     pub trino_auth: Option<String>,
     pub trino_schema: Option<String>,
     pub trino_catalog: Option<String>,
+
+    // audit log vars
+    pub audit_log_target: Option<String>,
+    pub audit_log_target_username: Option<String>,
+    pub audit_log_target_password: Option<String>,
+    pub audit_log_target_tls_verify: bool,
+    pub audit_log_target_headers: HashMap<String, String>,
```
de-sh, line 128: This could just as well be an [`http::HeaderMap`](https://docs.rs/http/latest/http/header/struct.HeaderMap.html)

server/src/handlers/http/middleware.rs
```diff
@@ -164,6 +171,31 @@ where
...
+                .unwrap_or("unknown"),
+                "datetime": Utc::now(),
+                "action":self.action,
+                "Actor":{
+                    "type": &req
+                    .headers()
+                    .get("user-agent")
+                    .and_then(|value| value.to_str().ok())
+                    .unwrap_or("unknown"),
+                    "id": "user123"
```
de-sh, line 190: Why is this hardcoded, and what is the intention behind using user-agent as type?

server/src/handlers/http/middleware.rs
```diff
@@ -164,6 +171,31 @@ where
...
+                },
+                  "ip-address":&req
```
de-sh, line 192: Are you intending to use the addr of the client?

server/src/handlers/http/middleware.rs
```diff
@@ -164,6 +171,31 @@ where
         /* ## Section end */
 
         let auth_result: Result<_, Error> = (self.auth_method)(&mut req, self.action);
+        let body = json!([
```
de-sh, line 174: I will work on the Schema, will get back to you in a week's time

Anirudhxx: Hey @de-sh, just following up. Did you get a chance to work on this?

de-sh: Hey, sorry that this got lost, here is what I have been able to come up with:
```json
{
  "audit_id": "String: uuid",
  "timestamp": "String: Datetime",
  "actor": {
    "type": "Enum: user/sync",
    "user_id": "String: username",
    "authorization_method": "Enum: basic/oauth/jwt"
  },
  "request": {
    "method": "Enum: POST/GET/PUT",
    "url": "String: url",
    "headers": {
      "key": "value"
    },
    "body": "String: json",
    "protocol": "Enum: kafka/http"
  },
  "action": {
    "type": "Enum: create/delete/update",
    "target": "Enum: s3/hottier/local-store/user-list",
    "resource": "Enum: object/user"
  },
  "response": {
    "code": int,
    "body": "String",
    "headers": {
      "key": "value"
    }
  }
}
```

de-sh:
> This most likely add a performance penalty to our server and should be moved elsewhere (Looking for some suggestions)

I feel like the right way to do this would involve a dedicated thread that handles the serialization and `reqwest` part of your code. This since audit logging shouldn't impact normal functioning of request in any-ways and hence should be done, removed from the data flow path. I am exploring the feasibility of using a specialized `tracing`  subscriber, to do this for us, similar to how we log to stdout now.

Anirudhxx:
> I am exploring the feasibility of using a specialized tracing subscriber, to do this for us, similar to how we log to stdout now and how otel loggers work.

@de-sh I'll implement a audit macro and replace it with the current reqwest code and create a new tracing subscriber which will look for these audit spans and send the audit log to the relevant parseable server. lmk if this works?

nikhilsinhaparseable:
@de-sh can we add below data points in the ingestion handler -
1. timestamp of the incoming request
2. HTTP status code of the response
3. if not 200 OK, message as well

if we send this to a stream on a different Parseable instance, we can atleast verify if request reached Parseable server and it responded with 200 OK or did it reject.

de-sh:
@Anirudhxx do let us know if you are going to continue or abandoning this PR

Anirudhxx:
@de-sh I will work on this today.

nitisht:
hey @Anirudhxx this is needed for a production deployment, so we've been working on this in parallel. Closing this due to no response. Thank you for the effort, unfortunately we couldn't get this in.

PR #1000, opened 2024-11-21
refactor: restructure as a library

de-sh:
Fixes #XXXX.

### Description

<hr>

This PR has:


de-sh:
The codepath being pointed out by Deepsource is generated and can be ignored

PR #1011, opened 2024-11-28
feat: schema detection for k8s events

nikhilsinhaparseable:
framework setup for schema detection at the time of ingestion sample json added for k8s events
formats file added that holds the list of sample jsons, schema type of all known log sources

server loads the known schemas at the initialization at the time of incoming events,
it checks if schema of incoming events match with any of the known schema if yes, it adds `schema_type` to the stream info

custom flattening is required before storing the schema and ingesting to parseable for those events which have hierarchical structure


nitisht:
closing in favour of #1248

PR #1016, opened 2024-12-02
fix: skip null data type fields

nikhilsinhaparseable:
skip the schema generation with `Null` data type fields 
fixes the issue when first event has null data type and field has valid data after the initial schema is persisted
fixes the schema merge issue for such cases

### DIFF
--- a/src/event/format/json.rs
+++ b/src/event/format/json.rs
@@ -86,6 +86,7 @@ impl EventFormat for Event {
                     infer_schema
                         .fields
                         .iter()
+                        .filter(|field| *field.data_type() != DataType::Null)
                         .cloned()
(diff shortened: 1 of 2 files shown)


src/event/format/json.rs
```diff
@@ -86,6 +86,7 @@ impl EventFormat for Event {
                     infer_schema
                         .fields
                         .iter()
+                        .filter(|field| *field.data_type() != DataType::Null)
```
de-sh, line 89:
```suggestion
                        .filter(|field| !field.data_type().is_null())
```

de-sh: LGTM otherwise

de-sh (approved):

PR #1022, opened 2024-12-05
refactor: parts of hottier

de-sh:
Fixes #XXXX.

### Description

- Uses `FuturesUnordered` to manage async tasks
- Reuses `get_stream_path_for_date` where possible
- Refactors `human_size` parsing
- Streams through results instead of collecting
- Stores sizes as integers in memory

<hr>

This PR has:


src/utils/human_size.rs
```diff
@@ -0,0 +1,59 @@
...
+
+pub fn serialize<S>(bytes: &u64, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+{
+    let human_readable = bytes_to_human_size(*bytes);
+    serializer.serialize_str(&human_readable)
+}
+
+pub fn deserialize<'de, D>(deserializer: D) -> Result<u64, D::Error>
```
de-sh, line 53: NOTE: this now returns B, KiB, etc instead of just Bytes. Could break things downstream if it only expects what came earlier

nikhilsinhaparseable:
@de-sh there is a change in behaviour of the PUT /hottier API
request body expected - 
before change: `{"size":"10737418240"}`
after change: `{"size":"10 GiB"}`

even the GET /hottier API is failing with error - 
```Serde(Error("Could not parse given size", line: 1, column: 42))```

can you please verify that there should be no behavioural change in the existing APIs.

nikhilsinhaparseable (approved):
looks good

PR #1024, opened 2024-12-06
refactor: utils/time parsing

de-sh:
Fixes #XXXX.

`TimeRange` is a new type to de-serialize strings into start and end `Datetime` and store it for representing a time range elsewhere in code. Currently usage is mainly in query.

### Description

<hr>

This PR has:


nitisht (commented):
Since we're refactoring - can you add comments as well?

src/handlers/airplane.rs
```diff
@@ -200,38 +202,40 @@ impl FlightService for AirServiceImpl {
...
-                let ingester_metadatas = get_ingestor_info()
-                    .await
-                    .map_err(|err| Status::failed_precondition(err.to_string()))?;
-                let mut minute_result: Vec<RecordBatch> = vec![];
-
-                for im in ingester_metadatas {
-                    if let Ok(mut batches) = run_do_get_rpc(im, out_ticket.clone()).await {
-                        minute_result.append(&mut batches);
-                    }
+        let event = if send_to_ingester(
```
de-sh, line 209: the following is not my change, it is a side effect of cargo fmt

nitisht (approved):

PR #1027, opened 2024-12-08
fix: convert all number data types to float

nikhilsinhaparseable:
server checks if event has any number (all ints and floats) 
then update the data type of all numbers to Float64

### DIFF
--- a/src/event/format/json.rs
+++ b/src/event/format/json.rs
@@ -185,7 +185,9 @@ fn valid_type(data_type: &DataType, value: &Value) -> bool {
         DataType::Boolean => value.is_boolean(),
         DataType::Int8 | DataType::Int16 | DataType::Int32 | DataType::Int64 => value.is_i64(),
         DataType::UInt8 | DataType::UInt16 | DataType::UInt32 | DataType::UInt64 => value.is_u64(),
-        DataType::Float16 | DataType::Float32 | DataType::Float64 => value.is_f64(),
+        DataType::Float16 | DataType::Float32 | DataType::Float64 => {
+            value.is_f64() || value.is_i64() || value.is_u64()
+        }
         DataType::Utf8 => value.is_string(),
         DataType::List(field) => {
             let data_type = field.data_type();
--- a/src/event/format/json.rs
+++ b/src/event/format/json.rs
@@ -185,9 +185,9 @@ fn valid_type(data_type: &DataType, value: &Value) -> bool {
         DataType::Boolean => value.is_boolean(),
         DataType::Int8 | DataType::Int16 | DataType::Int32 | DataType::Int64 => value.is_i64(),
         DataType::UInt8 | DataType::UInt16 | DataType::UInt32 | DataType::UInt64 => value.is_u64(),
-        DataType::Float16 | DataType::Float32 | DataType::Float64 => {
-            value.is_f64() || value.is_i64() || value.is_u64()
-        }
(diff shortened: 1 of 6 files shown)


src/event/format/json.rs
```diff
@@ -184,7 +184,9 @@ fn valid_type(data_type: &DataType, value: &Value) -> bool {
         DataType::Boolean => value.is_boolean(),
         DataType::Int8 | DataType::Int16 | DataType::Int32 | DataType::Int64 => value.is_i64(),
         DataType::UInt8 | DataType::UInt16 | DataType::UInt32 | DataType::UInt64 => value.is_u64(),
-        DataType::Float16 | DataType::Float32 | DataType::Float64 => value.is_f64(),
+        DataType::Float16 | DataType::Float32 | DataType::Float64 => {
+            value.is_f64() || value.is_i64() || value.is_u64()
+        }
```
de-sh, line 189:
```suggestion
        DataType::Float16 | DataType::Float32 | DataType::Float64 => value.is_number(),
```

src/event/format/mod.rs
```diff
@@ -204,13 +204,41 @@ pub fn override_timestamp_fields(
...
+                        field.name(),
+                        DataType::Float64,
+                        field.is_nullable(),
+                    ))
+                } else {
+                    field.clone()
+                }
+            })
+            .collect::<Vec<Arc<Field>>>(),
+    ))
```
de-sh, line 231:
```suggestion
let fields = schema
        .fields()
        .iter()
        .map(|field| {
            if field.data_type().is_numeric() {
                let field = Field::new(field.name(), DataType::Float64, field.is_nullable());
                Arc::new(field)
            } else {
                field.clone()
            }
        })
        .collect::<Vec<Arc<Field>>>();

    Arc::new(Schema::new(fields))
```

src/event/format/mod.rs
```diff
@@ -204,13 +204,41 @@ pub fn override_timestamp_fields(
     Arc::new(Schema::new(updated_fields))
 }
 
+pub fn override_num_fields_from_schema(schema: Arc<Schema>) -> Arc<Schema> {
```
de-sh, line 207:
```suggestion
/// All number fields from inferred schema are forced into Float64
pub fn override_num_fields_from_schema(schema: Arc<Schema>) -> Arc<Schema> {
```

src/event/format/mod.rs
```diff
@@ -204,6 +204,24 @@ pub fn override_timestamp_fields(
     Arc::new(Schema::new(updated_fields))
 }
 
+/// All number fields from inferred schema are forced into Float64
+pub fn override_num_fields_from_schema(schema: Vec<Arc<Field>>) -> Vec<Arc<Field>> {
+    schema
+        .iter()
+        .map(|field| {
+            if field.data_type().is_numeric() && field.data_type() != &DataType::Float64 {
```
de-sh, line 212:
```suggestion
            if field.data_type().is_numeric() {
```

Effect is the same, slight cost benefit from the other, but higher mental load to read

de-sh: This is just a suggestion, we can make do with a note comment atop this

de-sh (approved):

src/event/format/mod.rs
de-sh: This is just a suggestion, we can make do with a note comment atop this

PR #1035, opened 2024-12-16
feat: DEPRECATE caching

de-sh:
Fixes #XXXX.

### Description

- Removes all code that manages caching, be it query cache or local cache for quick query.
- Responds to cache management API calls with `410 GONE` HTTP response.

<hr>

This PR has:


nitisht (approved):
LGTM

PR #1037, opened 2024-12-18
fix: update numeric fields data type in storage schema before merge

nikhilsinhaparseable:
this handles scenario where stream has existing data 
and event is ingested with additional fields where
infer schema differs from storage schema

server updates the storage schema before merge with infer schema and persist to storage


nitisht:
Closing for now, we'll reassess the approach we want to take for this issue

PR #1038, opened 2024-12-18
Revert "fix: convert all number data types to float"

nitisht:
Reverts parseablehq/parseable#1027


nitisht:
Merging because locally quest seems to pass

PR #1043, opened 2024-12-19
feat: Custom Flattening for OTEL logs, metrics and traces

nikhilsinhaparseable:
- add proto files for metrics and trace
- add compiled rust files for metrics and trace protobuf files
- add separate handlers for OTEL logs, metrics and traces
- custom flattening added for OTEL logs, metrics and traces

    use endpoints
- `/v1/logs` for OTEL logs
- `/v1/metrics` for OTEL metrics
- `/v1/traces` for OTEL traces

    add custom header X-P-Log-Source when using endpint `api/v1/ingest`
- `otel-logs` for OTEL logs
- `otel-metrics` for OTEL metrics
- `otel-traces` for OTEL traces


de-sh:
@nikhilsinhaparseable can you document the code with expectations similar to this: https://github.com/parseablehq/parseable/pull/1024/files#diff-c45a9621970d114a8f2791b564a0a3dd8bf7825dcf549fae486a98032723c12eR41-R57

Please also move files into related module e.g. `otel_logs.rs` ~> `otel/logs.rs`

As for fixing the clippy, this issue is caused by including generated code into the repo, which is a bad practice, but given our circumstance, as a short term fix, we can use this:
```
#[allow(clippy::all)]
pub mod proto;
```

Cargo.toml
```diff
@@ -117,11 +117,6 @@ zip = { version = "2.2.0", default-features = false, features = ["deflate"] }
 url = "2.4.0"
 prost-build = "0.13.3"
```
de-sh, line 119:
```suggestion

[dev-dependencies]
maplit = "1.0"
rstest = "0.23.0"
arrow = "53.0.0"
```

de-sh:
We can consider using https://crates.io/crates/opentelemetry-proto

nikhilsinhaparseable:
@de-sh removed the pre-compiled and the protobuf files.
I am using the crate you suggested.
Please review again.

src/otel/metrics.rs
```diff
@@ -0,0 +1,483 @@
...
+            data_point_json.insert(key.clone(), value.clone());
+        }
+    }
+    data_points_json
+}
+
+/// this function performs the custom flattening of the otel metrics
+/// and returns a `Vec` of `BTreeMap` of the flattened json
+pub fn flatten_otel_metrics(body: &Bytes) -> Vec<BTreeMap<String, Value>> {
+    let body_str = std::str::from_utf8(body).unwrap();
```
de-sh, line 390: There looks like a possibility this could panic 😅

Cargo.toml
```diff
@@ -121,7 +122,6 @@ prost-build = "0.13.3"
 maplit = "1.0"
 rstest = "0.23.0"
 arrow = "53.0.0"
```
de-sh, line 124:
```suggestion
arrow = "53.0.0"

```

nitisht:
Is this ready now for merge?

src/event/format/json.rs
```diff
@@ -52,8 +52,17 @@ impl EventFormat for Event {
         static_schema_flag: Option<&String>,
         time_partition: Option<&String>,
         schema_version: SchemaVersion,
+        log_source: &str,
```
de-sh, line 55: We should move away from using str given how wide the possibilities are, please consider using enums similar to what we are doing with `SchemaVersion` above, it will also make things easier to read

src/otel/otel_utils.rs
```diff
@@ -0,0 +1,162 @@
...
+    let mut value_json: BTreeMap<String, Value> = BTreeMap::new();
+    match value {
+        OtelValue::StringValue(str_val) => {
+            value_json.insert(key.to_string(), Value::String(str_val));
+        }
+        OtelValue::BoolValue(bool_val) => {
+            value_json.insert(key.to_string(), Value::Bool(bool_val));
+        }
+        OtelValue::IntValue(int_val) => {
+            value_json.insert(key.to_string(), Value::String(int_val.to_string()));
```
de-sh, line 34: Can you remind me why we are doing int to string, instead of number?

nikhilsinhaparseable: otel collector sends int value as string, but now it can be serialised as int, hence, i think we can now change it to number, let me test it out

de-sh (approved):

nitisht:
@nikhilsinhaparseable can you resolve the conflicts

PR #1044, opened 2024-12-19
fix: dynamic schema inference casts strings as timestamp

de-sh:
Fixes #XXXX.

### Description

- Fixes timestamp field detection in dynamic schema inference.
- Operates with the same schema inference step in detect schema API.
- Improves performance of schema inference step during ingestion by streamlining steps.

<hr>

This PR has:

### DIFF
--- a/src/event/format/mod.rs
+++ b/src/event/format/mod.rs
@@ -241,18 +241,16 @@ fn update_schema_from_logs(
             if !existing_field_names.contains(&field.name())
                 && field.data_type() == &DataType::Utf8
                 && TIME_FIELD_PREFIXS.iter().any(|p| field.name().contains(p))
-            {
-                if log_records
+                && log_records
                     .iter()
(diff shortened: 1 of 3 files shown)


de-sh:
To test this we should update quest to make calls on schema detection API @nikhilsinhaparseable

src/event/format/mod.rs
```diff
@@ -201,88 +211,76 @@ pub fn override_timestamp_fields(
...
-                            // Update the field's data type to Timestamp
-                            return Field::new(
-                                field.name().clone(),
-                                DataType::Timestamp(TimeUnit::Millisecond, None),
-                                true,
-                            );
-                        }
-                    }
-                }
+            if field.name() == field_name && field.data_type() == &DataType::Utf8 {
```
de-sh, line 274: what if the field has a non-utf8 type?

PR #1048, opened 2024-12-22
BugFix: Fixed query auth

parmesant:
Auth flow for query and permission assignment for `ListStream` changed

Fixes #XXXX.

### Description

This PR fixes a flaw in the auth flow for the `query` API
It also changes the Permission for the action `ListStream` from `Unit` to `Stream` which will limit the response of the API to only the streams that the user is allowed to see

<hr>

This PR has:


nikhilsinhaparseable (commented):
fix is validated, @de-sh  please review the code once

src/handlers/http/logstream.rs
```diff
@@ -85,16 +89,27 @@ pub async fn delete(req: HttpRequest) -> Result<impl Responder, StreamError> {
     Ok((format!("log stream {stream_name} deleted"), StatusCode::OK))
 }
 
-pub async fn list(_: HttpRequest) -> impl Responder {
-    //list all streams from storage
+pub async fn list(req: HttpRequest) -> Result<impl Responder, StreamError> {
+    let key = extract_session_key_from_req(&req)
+        .map_err(|err| StreamError::Anyhow(anyhow::Error::msg(err.to_string())))?;
```
de-sh, line 94:
```suggestion
        .map_err(|err|  anyhow::Error::msg(err.to_string()))?;
```

src/handlers/http/logstream.rs
```diff
@@ -85,16 +89,27 @@ pub async fn delete(req: HttpRequest) -> Result<impl Responder, StreamError> {
...
+
+    // list all streams from storage
     let res = CONFIG
         .storage()
         .get_object_store()
         .list_streams()
         .await
-        .unwrap();
+        .unwrap()
```
de-sh, line 102:
```suggestion
        .unwrap()
```
??

src/handlers/http/query.rs
```diff
@@ -156,38 +156,41 @@ pub async fn create_streams_for_querier() {
...
-                authorized = true;
-                break;
-            }
-            Permission::StreamWithTag(Action::Query, ref stream, tag)
-                if stream == table_name || stream == "*" =>
-            {
-                authorized = true;
-                if let Some(tag) = tag {
-                    tags.push(tag)
+    for table_name in tables.iter() {
```
de-sh, line 161: I'd suggest reverting changes here and moving it to the call site

de-sh: e.g:
```
       for stream_name in streams {
            authorize_and_set_filter_tags(&mut query, &permissions, &stream_name).map_err(
                |_| Status::permission_denied("User Does not have permission to access this"),
            )?;
        }
```

de-sh (changes requested):

PR #1049, opened 2024-12-22
Enhancement: Re-design alerts

parmesant:
### Description

Alerts used to be linked to a logstream and were based on an expression based logic With this PR, alerts will be independent of logstreams and will be based on a SQL query The given threshold config will get evaluated against the result of the query and the target config will be used to send out notifications

<hr>

**TODO:**

- [ ] Unit tests
- [ ] Merge target notification task with Alerts

This PR has:


nitisht:
Please address the conflicts

PR #1056, opened 2024-12-25
feat: clap_completion

TwistingTwists:
- autocompletion for the cli

No issue created yet.

### Description

Adding auto completion for the cli - to ease developer UX.

<hr>

This PR has:

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -44,7 +44,7 @@ use std::string::String as KafkaSslProtocol;
 pub const DEFAULT_USERNAME: &str = "admin";
 pub const DEFAULT_PASSWORD: &str = "admin";
 
-#[derive(Parser)]
+#[derive(Parser, Debug)]
 #[command(
     name = "parseable",
     bin_name = "parseable",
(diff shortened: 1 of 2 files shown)


TwistingTwists:
Here is the [DRAFT PR ](https://github.com/parseablehq/parseable/pull/1056/files):
Questions:

1. it changes the [CONFIG](https://github.com/parseablehq/parseable/pull/1056/files#diff-80bcaad8240646d1b9ce48bf64176900a29f019d75efbdb1df14916da07a7ffaR41) struct. Since, CONFIG is used almost everywhere in code, it'll cause lots of small changes across codebase. [The changes will look like this](https://github.com/parseablehq/parseable/pull/1056/files#diff-3fb0efd6cb1750d42bfe3c2acc60a324089286d5ca75cf24c5eeb362555d1cc7R155-R158)

2. Currently the new enum `ParseableCommand` delegates [the methods related to Config struct](https://github.com/parseablehq/parseable/pull/1056/files#diff-80bcaad8240646d1b9ce48bf64176900a29f019d75efbdb1df14916da07a7ffaR176-R260) in impl ParseableCommand.
I think this could be handled by defining a trait .

Is there an alternate approach I haven't considered that you would like me to try?

de-sh:
@TwistingTwists can you please resolve conflicts and update your branch?

TwistingTwists:
Yup. will do

TwistingTwists:
Unresolved Questions about this PR: 

1. currently, the cli exits after generations the completions from the `pb completions` command . Since this is inside `Config` struct, we don't return anything. This is not ideal. 

if there is an alternate you would like me to try, please let me know. 
cc @de-sh

src/cli.rs
```diff
@@ -73,13 +73,88 @@ Join the community at https://logg.ing/community.
...
+// formatting without disabling rustfmt for the entire struct.
+#[doc = r#"Print a command-line-completion script
+
+Apply it by running one of these:
+
+- Bash: `source <(pb util completion bash)`
+- Fish: `pb util completion fish | source`
+- Nushell:
+     ```nu
+     pb util completion nushell | save "completions-pb.nu"
```
de-sh, line 97: `pb` is a different cli tool, a client to the parseable(server) actually. 

Can we have a argument/flag similar to `--help` that when set will print the completions for the specific shell. i.e.
```sh
parseable --shell-completion-config nushell | save "parseable-completions.nu"
```

src/cli.rs
```diff
@@ -73,13 +73,88 @@ Join the community at https://logg.ing/community.
...
+    Elvish,
+    Fish,
+    // Nushell,
+    PowerShell,
+    Zsh,
+}
+
+impl ShellCompletion {
+    pub fn generate(&self, cmd: &mut clap::Command) -> Vec<u8> {
+        use clap_complete::generate;
```
de-sh, line 127: The `clap_complete` is missing?

src/cli.rs
TwistingTwists: This is also a good way to do this, rather than adding a command. I'll implement this.

PR #1057, opened 2024-12-27
feat: conditionalize generic flattening and type casting for streams v1+

de-sh:
Fixes #XXXX.

### Description

- For streams created after the v1 update, generic flattening will be applied.
- All numbers will be stored as float64.
- All string fields with name "date"/"time" that can be parsed as timestamp will be stored as timestamp.

<hr>

This PR has:

### DIFF
--- a/src/handlers/http/ingest.rs
+++ b/src/handlers/http/ingest.rs
@@ -786,8 +786,8 @@ mod tests {
         assert_eq!(rb.num_rows(), 4);
         assert_eq!(rb.num_columns(), 7);
         assert_eq!(
-            rb.column_by_name("a").unwrap().as_int64_arr().unwrap(),
-            &Int64Array::from(vec![Some(1), Some(1), Some(1), Some(1)])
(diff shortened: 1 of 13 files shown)


src/handlers/http/ingest.rs
```diff
@@ -84,7 +84,7 @@ pub async fn ingest_internal_stream(stream_name: String, body: Bytes) -> Result<
             tags: String::default(),
             metadata: String::default(),
         };
-        event.into_recordbatch(&schema, None, None)?
+        event.into_recordbatch(&schema, None, None, SchemaVersion::V0)?
```
nitisht, line 87: Let's add a comment explaining what is V0 vs V1 (and what is the different)

nikhilsinhaparseable (dismissed):
looks good

nitisht (changes requested):
Please add detailed comments especially in places we're checking for versions. Also we need comments on what is `v0`

PR #1058, opened 2024-12-30
fix: hierarchical json flattening restriction

nikhilsinhaparseable:
get the level of hierarchy from the json
perform generic flattening only if level of nesting is <=4

### DIFF
--- a/src/handlers/http/modal/utils/ingest_utils.rs
+++ b/src/handlers/http/modal/utils/ingest_utils.rs
@@ -23,13 +23,12 @@ use anyhow::anyhow;
 use arrow_schema::Field;
 use bytes::Bytes;
 use chrono::{DateTime, NaiveDateTime, Utc};
-use itertools::Itertools;
 use serde_json::Value;
 
 use crate::{
     event::{
+        self,
         format::{self, EventFormat},
-        Event,
     },
     handlers::{
         http::{ingest::PostError, kinesis},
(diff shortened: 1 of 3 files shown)


src/utils/json/flatten.rs
```diff
@@ -273,52 +274,83 @@ pub fn flatten_array_objects(
...
 /// 1. `{"a": 1}` ~> `[{"a": 1}]`
 /// 2. `[{"a": 1}, {"b": 2}]` ~> `[{"a": 1}, {"b": 2}]`
 /// 3. `[{"a": [{"b": 1}, {"c": 2}]}]` ~> `[{"a": {"b": 1)}}, {"a": {"c": 2)}}]`
-/// 3. `{"a": [{"b": 1}, {"c": 2}], "d": {"e": 4}}` ~> `[{"a": {"b":1}, "d": {"e":4}}, {"a": {"c":2}, "d": {"e":4}}]`
-pub fn flatten_json(value: &Value) -> Vec<Value> {
+/// 4. `{"a": [{"b": 1}, {"c": 2}], "d": {"e": 4}}` ~> `[{"a": {"b":1}, "d": {"e":4}}, {"a": {"c":2}, "d": {"e":4}}]`
+/// 5. `{"a":{"b":{"c":{"d":{"e":["a","b"]}}}}}` ~> returns error - heavily nested, cannot flatten this JSON
+pub fn flatten_json(value: &Value) -> Result<Vec<Value>, anyhow::Error> {
+    if has_more_than_four_levels(value, 1) {
+        return Err(anyhow!("heavily nested, cannot flatten this JSON"));
```
de-sh, line 288: Why error out? Why not just not flatten and keep the type as is?

src/utils/json/flatten.rs
```diff
@@ -273,52 +274,83 @@ pub fn flatten_array_objects(
...
+        _ => Ok(vec![value.clone()]),
+    }
+}
+
+/// recursively checks the level of nesting for the serde Value
+/// if Value has more than 4 levels of hierarchy, returns true
+/// example -
+/// 1. `{"a":{"b":{"c":{"d":{"e":["a","b"]}}}}}` ~> returns true
+/// 2. `{"a": [{"b": 1}, {"c": 2}], "d": {"e": 4}}` ~> returns false
+fn has_more_than_four_levels(value: &Value, current_level: usize) -> bool {
```
de-sh, line 342: This would just add compute cycles to ingestion without helping in any other way, do we need this?

src/utils/json/flatten.rs
de-sh: i.e. `{"a": [{"b":{"c":{"d":["e","f"]}}}, "c": {}]}` ~> `[{"a":{"b":{"c":{"d":["e","f"]}}}}, {"a":{"c": {}}}]`

src/handlers/http/modal/utils/ingest_utils.rs
```diff
@@ -73,61 +72,174 @@ pub async fn push_logs(
...
+            )
+            .await?;
+        } else {
+            let data = convert_array_to_object(
+                body_val,
+                None,
+                None,
+                custom_partition.as_ref(),
+                schema_version,
+            )?;
```
de-sh, line 100: Would the following fix the issue?
```
let data = if time_partition.is_some() || custom_partition.is_some() {
        convert_array_to_object(
            body_val,
            time_partition.as_ref(),
            time_partition_limit,
            custom_partition.as_ref(),
            schema_version,
        )?
    }  else {
        vec![body_val]
    };
```

nikhilsinhaparseable: I need to test out all the scenarios before I can confirm the changes, as discussed, lets take the optimisation of this one as separate issue.

de-sh (approved):

PR #1063, opened 2025-01-03
feat: audit logging

### LINKED ISSUE #765: feat: audit logging for Parseable

de-sh:
Fixes #765.

### Description
Audit logs are sent to another Parseable instance everytime a request is received over http or when new logs are received over kafka.

While we did explore the option, we chose to perform audit logging without relying on tracing, given the complexity of the setup.

In the PR we have been as careful as possible to not cause unnecessary allocations if we don't end up logging anyways.

<hr>

This PR has:


src/audit.rs
```diff
@@ -0,0 +1,255 @@
...
+            .join("/api/v1/ingest")
+        {
+            Ok(url) => url,
+            Err(err) => {
+                eprintln!("Couldn't setup audit logger: {err}");
+                return None;
+            }
+        };
+
+        let client = Arc::new(reqwest::Client::new());
```
hippalus, line 65: I would use a builder pattern here with these client configs (due to https://github.com/seanmonstar/reqwest/issues/1454, https://github.com/hyperium/hyper/issues/3071  issues.

I saw multiple Client::new() in other places (unrelated to this PR) creating new clients per request. I would consider having a ParseableHttpClient wrapper to share Arc<Client> instances. Nice to have reqwest-middleware (https://crates.io/crates/reqwest-middleware) for retrying and logging middleware.


   ```rust
   ClientBuilder::new()
       .connect_timeout(Duration::from_secs(10)) // may configurable
       .timeout(Duration::from_secs(5)) // may configurable
       .gzip(true)  // Nice to have for compression
       .use_rustls_tls()  // For TLS implementation
       .http1_only()  // To avoid HTTP/2 issues
       .build()?`

src/kafka.rs
```diff
@@ -274,8 +275,24 @@ pub async fn setup_integration() {
...
+        log_builder.set_request("", "", "Kafka", []);
+        log_builder.set_stream_name(curr.topic());
+
+        let Err(err) = ingest_message(curr).await else {
+            log_builder.set_response(200, "");
+            continue;
+        };
+        log_builder.set_response(500, &err);
+        error!("Unable to ingest incoming kafka message- {err}");
+        log_builder.send().await;
```
hippalus, line 296: With the new Kafka implementation, this could become a bottleneck since we're making HTTP calls for each message. Would it be possible to add batch support to the endpoint to make this more efficient?

src/kafka.rs
```diff
@@ -274,8 +275,24 @@ pub async fn setup_integration() {
...
-        if let Err(err) = ingest_message(curr).await {
-            error!("Unable to ingest incoming kafka message- {err}")
-        }
+        // Constructs a log for each kafka request
+        let mut log_builder = AuditLogBuilder::default();
+        log_builder.set_actor(
+            CONFIG.parseable.kafka_host.as_deref().unwrap_or(""),
+            "Kafka Client",
+            "",
+            "",
```
hippalus, line 284: The builder pattern could be more elegant here.
 I would:

-  Chain the builder calls like `.set_actor().set_request().set_stream()` and be sure that prefixes like "set", "with", etc compatible with other builders in the codebase.

-  Handle optional or default values inside builder methods
-  Add precondition checks in build()/send() if validation logics  requires

```rust
AuditLogBuilder::default()
    .set_actor(kafka_host.unwrap_or_default(), "Kafka Client")
    .set_request_type("Kafka")
    .set_stream(curr.topic())
    .set_response(status, err.as_str())
    .send()
    .await

**Review by hippalus — COMMENTED** [review]
LGTM. Few comments

File: src/kafka.rs (hunk 2 above)
Comment (de-sh) [reply src/kafka.rs:296]: Understood, will look into this

**Comment by nitisht** [comment]
proposed format:

```
{
  "parseable_server_version": "1.6.3",
  "parseable_server_deploymentId": "01JGR7PS9K28JRKNS4V6F2WNAK",
  "audit_version": "V1",
  "audit_generation_timestamp": "2025-01-04T14:16:28.483+05:30",
  "audit_id": "01JGR7X443QX5N45B4ZPNN5QEG",
  "actor_authorizationMethod": "Basic Auth",
  "actor_remote_host": "127.0.0.1",
  "actor_username": "admin",
  "actor_user_agent": "PostmanRuntime/7.43.0",
  "request_headers_accept": "*/*",
  "request_headers_accept-encoding": "gzip, deflate, br",
  "request_headers_connection": "keep-alive",
  "request_headers_content-length": "100",
  "request_headers_content-type": "application/json",
  "request_headers_host": "localhost:8000",
  "request_headers_postman-token": "6a1bd4ae-0395-4e5c-90df-b7b5105bf0b5",
  "request_method": "POST",
  "request_start_time": "2025-01-04T14:16:28.480+05:30",
  "request_ end_time": "2025-01-04T14:16:28.483+05:30",
  "request_path": "/api/v1/ingest",
  "request_protocol": "http",
  "request_body_size": "1234",
  "response_status_code": 400,
  "response_error_message": "",
  "response_body_size": "1234",
  "stream": "app",
}
```

src/audit.rs
de-sh: Something like this: https://github.com/parseablehq/parseable/commit/a6adeb06442e7dcf868944a22c3ee013349335a3?

hippalus: Kind of yes but // The connect_timeout (10s) exceeds the request timeout (5s), creating a potential issue:
// If connection establishment takes 7s, the entire request would time out before the 
// connection is even established. For better reliability, connect_timeout should be shorter 
// than the request timeout, like:
// .connect_timeout(Duration::from_secs(3))
// .timeout(Duration::from_secs(10))

hippalus: Thanks  LGTM.  I am adding a clearer example. 


   ```rust
static HTTP_CLIENT: Lazy<Client> = Lazy::new(|| {
    ClientBuilder::new()
        // Maximum time allowed for establishing a TCP connection (3s)
        // - Should be shorter than the overall request timeout
        // - Helps fail fast when network connectivity issues occur 
        // - Typical values range from 3-5 seconds
        .connect_timeout(Duration::from_secs(3))

        // Maximum time for the entire HTTP operation (10s)
        // - Includes connection time, request transmission, and response reception
        // - Should be longer than connect_timeout
        .timeout(Duration::from_secs(10))
        .gzip(true)

        // Enable Brotli compression
        // - Since the proposed format   "request_headers_accept-encoding": "gzip, deflate, br",
        .brotli(true)

        // Use Rustls TLS implementation
        .use_rustls_tls()

        // How long to keep unused connections (90s)
        // - Balances resource usage with connection reuse
        // - Helps reduce connection establishment overhead
        .pool_idle_timeout(Duration::from_secs(90))

        // Maximum concurrent connections per host
        // - 32 connections suitable for most applications but consider loads. default is usize::MAX
        // - Prevents excessive resource consumption. 
        .pool_max_idle_per_host(32) 

        // Build the client
        .build()
        .expect("HTTP client initialization failed")
});`

src/kafka.rs
de-sh: We will be taking this up in a separate PR

src/audit.rs
de-sh: We will be taking this up in a separate PR

PR #1070, opened 2025-01-06
refactor: use `bool`s for flags

de-sh:
Fixes #XXXX.

### Description

Only "true" is valid, deserializing anything else is not allowed, if the field doesn't exist only then value is set to false, this is perfectly fine when reading from file and this will not impact previous behaviour since we also serialize only true as "true" and ignore false value(as we have been doing previously).

This improve perf by not having to pass along an entire `String`, and being able to make do with a bool, in memory.

<hr>

This PR has:


nikhilsinhaparseable (approved):
good to merge

PR #1071, opened 2025-01-06
refactor: single `reqwest::Client` for HTTP

de-sh:
Fixes #XXXX.

### Description

Use a single HTTP client to perform all normal outgoing HTTP requests from the parseable server.

Connection timeouts in 3s during setup, 10s per request and every 90s when not used/idling. Max 32 idle connections are allowed per host. gzip/brotli compression is enabled and can be used based on server(to which parseable connects) preference, rustls is used to perform TLS encryption (instead of native-tls) and only HTTP/1.1 connections are allowed (no HTTP/2)

This is different from the default client which has no timeouts, default pooling, no compression, and would support multiple TLS backends and HTTP versions.

<hr>

This PR has:


Cargo.toml
```diff
@@ -73,6 +73,8 @@ relative-path = { version = "1.7", features = ["serde"] }
 reqwest = { version = "0.11.27", default-features = false, features = [
```
hippalus, line 73: would be nice to have the latest version 0.12.x if there is no blocker issue.

de-sh: We will take this up in #1075

de-sh: `actix_http:latest` and `reqwest:v0.11+` are incompatible because one depends on different versions of `http`([ref1](https://github.com/actix/actix-web/blob/master/actix-http/Cargo.toml#L116), [ref2](https://github.com/seanmonstar/reqwest/blob/master/Cargo.toml#L103)).

nikhilsinhaparseable (approved):
looks good

nitisht:
@de-sh can you resolve the conflicts pls

PR #1080, opened 2025-01-07
feat: batch audit logs and ensure they are not lost

de-sh:
Fixes #XXXX.

### Description

Collects audit logs into batches, pushes as configured based on threshold of time/size and ensures logs are persisted into a file when the network is down. NOTE: Works in a near synchronous manner where logs will not be accepted if write to disk is blocked or when direct push to network is blocked, this may lead to a situation where requests will not respond properly as they try to send audit logs, but this is a known cost.

<hr>

This PR has:

## Summary by CodeRabbit

- **New Features**
  - Enhanced audit logging with a new `AuditLogger` for managing and sending logs, including batch processing and improved error handling.
  - Introduced configurable options for audit log batch size, flush intervals, and log storage directory via environment variables.
  
- **Refactor**
  - Streamlined logging components and shutdown procedures for reliable initialization and graceful termination of the audit logging system.


src/audit/logger.rs
```diff
@@ -0,0 +1,221 @@
...
+        logger.next_log_file_id = latest_log_file_id + 1;
+        if oldest_log_file_id != usize::MAX {
+            logger.oldest_log_file_id = oldest_log_file_id;
+        }
+
+        logger
+    }
+}
+
+impl AuditLogger {
```
hippalus, line 95: It would be nice to have graceful shutdown support to prevent log loss.

src/audit/logger.rs
```diff
@@ -0,0 +1,221 @@
...
+            .parseable
+            .audit_log_dir
+            .join(format!("{}.json", self.next_log_file_id));
+        let mut log_file = OpenOptions::new()
+            .create(true)
+            .truncate(true)
+            .open(log_file_path)
+            .await
+            .expect("Failed to open audit log file");
+        let buf = serde_json::to_vec(&logs_to_send).expect("Failed to serialize audit logs");
```
hippalus, line 124: Possibly it looks like it never happens, but unwrap() can cause panic in production. It would be nice to have proper error handling with context.

src/audit/logger.rs
```diff
@@ -0,0 +1,221 @@
...
+        log_file.write_all(&buf).await.unwrap();
```
hippalus, line 125: same

src/audit/logger.rs
```diff
@@ -0,0 +1,221 @@
...
+        let logs_to_send: Vec<AuditLog> = serde_json::from_reader(&mut oldest_file)?;
+        self.send_logs_to_remote(&logs_to_send).await?;
+
+        // Delete the oldest log file
+        std::fs::remove_file(oldest_file_path)?;
+
+        Ok(())
+    }
+
+    async fn send_logs_to_remote(&self, logs: &Vec<AuditLog>) -> Result<(), anyhow::Error> {
```
hippalus, line 163: why not anyhow::Result<

src/audit/logger.rs
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    /// Spawns a background task for periodic flushing of audit logs, if configured
+    pub async fn spawn_batcher(mut self) {
+        if self.log_endpoint.is_none() {
+            return;
+        }
+
+        // setup the audit log channel
+        let (audit_log_tx, mut audit_log_rx) = channel(0);
```
hippalus, line 192: why the initial size is 0 ?

hippalus (commented):
LGTM! The implementation looks solid, but I have some questions about the architectural approach:

Audit Log Architecture:
1. What drove the decision to use another Parseable instance for audit logs rather than local storage?
2. Do we have plans for a dedicated audit logging solution in the future?

Deployment Considerations:
For multi-node setups (e.g., 3 ingest + 1 query nodes):
1. What's the recommended log_endpoint configuration?
2. Which node should own audit log collection?
3. How do we prevent circular logging dependencies?

Performance & Scalability:
1. In high-traffic scenarios, how do we handle the additional network load from audit logging?
2. Have we considered the impact of inter-service dependencies on system reliability?

Would appreciate your insights on these points to better understand the architectural vision.

src/audit/logger.rs
```diff
@@ -0,0 +1,221 @@
...
+        // Try to construct the log endpoint URL by joining the base URL
+        // with the ingest path, This can fail if the URL is not valid,
+        // when the base URL is not set or the ingest path is not valid
+        let Some(url) = CONFIG.parseable.audit_logger.as_ref() else {
+            return logger;
+        };
+
+        logger.log_endpoint = url
+            .join("/api/v1/ingest")
+            .inspect_err(|err| eprintln!("Couldn't setup audit logger: {err}"))
```
hippalus, line 59: `tracing::error! `It would be nice

src/audit/logger.rs
de-sh: This should not be an issue in most places are serialization is not a problem for these types. But still we have expects to document why it is safe to assume it won't

src/audit/logger.rs
de-sh: This doubt makes sense, will add changes to handle this scenario

de-sh:
Thank you for the comments @hippalus, let me note that audit logging is not to be considered as beta yet. We are working on developing this as a feature that is still in development and will need a lot of work before we can consider it ready for production. 

With regards to your questions:

> LGTM! The implementation looks solid, but I have some questions about the architectural approach:
> 
> Audit Log Architecture:
> 
>     1. What drove the decision to use another Parseable instance for audit logs rather than local storage?

It is an industry practice to send audit logs to an external system, we felt dogfooding would be the least friction option when it comes to doing this.

> 
>     2. Do we have plans for a dedicated audit logging solution in the future?
> 

Parseable already fulfills most of this and we don't see any reason as to why we would build with anything different. Please let us know of any alternatives, we can document them and plan out how to build support. To make ourselves clear, logs should be sent to an instance outside of the cluster.

> 
> Deployment Considerations: For multi-node setups (e.g., 3 ingest + 1 query nodes):
> 
>     1. What's the recommended log_endpoint configuration?
> 
The log_endpoint should be a separate cluster/standalone instance of parseable.

>     2. Which node should own audit log collection?
> 

This should not matter in my opinion, in a multi-node setup the expectation is that a load-balancer should take the decision and logs should end up in the same stream no matter the node it is received at.

>     3. How do we prevent circular logging dependencies?
> 

Circular logs are not expected as pointed above.

> 
> Performance & Scalability:
> 
>     1. In high-traffic scenarios, how do we handle the additional network load from audit logging?
> 

We are only testing the feature and as of our current learnings, batching and writing to file on network slow-down as well as pushing to file when network is down seemed to be the right decision for us to handle high-frequency use-cases.

>     2. Have we considered the impact of inter-service dependencies on system reliability?
> 

What do you mean by "inter-service dependencies".

> 
> Would appreciate your insights on these points to better understand the architectural vision.

hippalus:
> Thank you for the comments @hippalus, let me note that audit logging is not to be considered as beta yet. We are working on developing this as a feature that is still in development and will need a lot of work before we can consider it ready for production.
> 
> With regards to your questions:
> 
> > LGTM! The implementation looks solid, but I have some questions about the architectural approach:
> > Audit Log Architecture:
> > ```
> > 1. What drove the decision to use another Parseable instance for audit logs rather than local storage?
> > ```
> 
> It is an industry practice to send audit logs to an external system, we felt dogfooding would be the least friction option when it comes to doing this.
> 
> > ```
> > 2. Do we have plans for a dedicated audit logging solution in the future?
> > ```
> 
> Parseable already fulfills most of this and we don't see any reason as to why we would build with anything different. Please let us know of any alternatives, we can document them and plan out how to build support. To make ourselves clear, logs should be sent to an instance outside of the cluster.
> 
> > Deployment Considerations: For multi-node setups (e.g., 3 ingest + 1 query nodes):
> > ```
> > 1. What's the recommended log_endpoint configuration?
> > ```
> 
> The log_endpoint should be a separate cluster/standalone instance of parseable.
> 
> > ```
> > 2. Which node should own audit log collection?
> > ```
> 
> This should not matter in my opinion, in a multi-node setup the expectation is that a load-balancer should take the decision and logs should end up in the same stream no matter the node it is received at.
> 
> > ```
> > 3. How do we prevent circular logging dependencies?
> > ```
> 
> Circular logs are not expected as pointed above.
> 
> > Performance & Scalability:
> > ```
> > 1. In high-traffic scenarios, how do we handle the additional network load from audit logging?
> > ```
> 
> We are only testing the feature and as of our current learnings, batching and writing to file on network slow-down as well as pushing to file when network is down seemed to be the right decision for us to handle high-frequency use-cases.
> 
> > ```
> > 2. Have we considered the impact of inter-service dependencies on system reliability?
> > ```
> 
> What do you mean by "inter-service dependencies".
> 
> > Would appreciate your insights on these points to better understand the architectural vision.

Thank you for the clarification @de-sh . I highlighted these questions because I couldn't find explicit documentation or configuration descriptions in the code addressing these architectural decisions. The implementation looks quite solid.

I'm particularly looking forward to seeing the Kafka connector integration. Regarding inter-service dependencies, as you've well explained, we can scale Parseable instances behind a load balancer for audit logging purposes. It's great to see this pathway being established.

IMHO : It would be nice to add more documentation around these architectural decisions directly in the code, making it easier for future contributors to understand the design choices. It would be nice to have to add docker-compose as an example.

PR #1083, opened 2025-01-08
make about utils public

nikhilsinhaparseable:


de-sh (approved):

src/storage/mod.rs
```diff
@@ -31,7 +31,7 @@ use std::fmt::Debug;
 mod azure_blob;
 mod localfs;
 mod metrics_layer;
-pub(crate) mod object_storage;
```
de-sh, line 34: this can be reverted

src/storage/mod.rs
```diff
@@ -31,7 +31,7 @@ use std::fmt::Debug;
...
+pub mod object_storage;
```
de-sh, line 34:
```suggestion
pub(crate) mod object_storage;
```
This change can be reverted

PR #1088, opened 2025-01-08
refactor: use `clap::Parser` for readability

de-sh:
Fixes #XXXX.

### Description

This has been in the works for long, finally figured out how to simplify the flow and get rid of the "junk" code involving unnecessary arg matching(totally my opinion), but this I believe is a readability win for the CLI and should allow us to work on further modularizing the codebase. It still needs a bit of testing to ensure parity with previous implementation.

<hr>

This PR has:


nikhilsinhaparseable (dismissed):
looks good

PR #1093, opened 2025-01-10
feat: API for date_bin query

parmesant:
new API to get the count(*) for the graph query
API - POST `/datebin`
sample request JSON -
```{"stream":"frontend", "startTime":"2025-01-08T10:10:00.000Z","endTime":"2026-01-08T23:20:00.000Z", "numBins":10}```
sample response JSON -
```
{
    "fields": [
        "date_bin_timestamp",
        "log_count"
    ],
    "records": [
        {
            "date_bin_timestamp": "2025-01-08T10:10:00+00:00",
            "log_count": 1250
        }
    ]
}
```
console sends the number of ticks to show in the graph based on the time range provided by the user
server fetches the stream.json and all manifest.jsons to get the count
avoid datafusion as query demands the count which is available in the catalog

leverage the same flow to get the count(*) queries without filters from query API
eg. `select count(*) from frontend` or `select count(*) as count from frontend`

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -23,7 +23,8 @@ use url::Url;
 
 use crate::{
     oidc::{self, OpenidConfig},
-    option::{validation, Compression, Mode}, storage::{AzureBlobConfig, FSConfig, S3Config},
+    option::{validation, Compression, Mode},
+    storage::{AzureBlobConfig, FSConfig, S3Config},
 };
 
 #[cfg(any(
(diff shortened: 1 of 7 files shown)


de-sh:
Let's work on this after merging #1100?

de-sh (changes requested):
Please consider: https://github.com/parmesant/parseable/pull/2

de-sh (approved):

PR #1094, opened 2025-01-10
refactor: signal handling is not related to http alone

de-sh:
Fixes #XXXX.

### Description

Graceful shutdown is not merely related to the web-server implementation and it should be kept separately from the code for the server

<hr>

This PR has:


nikhilsinhaparseable:
image
@de-sh the PR works well in linux but fails to compile in windows, please check the attached screenshot.

nikhilsinhaparseable (approved):
looks good

PR #1098, opened 2025-01-12
refactor: query request flow

de-sh:
Fixes #XXXX.

### Description

<hr>

This PR has:


de-sh:
Might be a good idea to learn from the suggested patterns in this PR, but for the long term I feel continuing to work on this PR is not worth it.

PR #1103, opened 2025-01-16
feat: rename API `datebin` to `counts`

de-sh:
Fixes #XXXX.

### Description

It makes more sense to call it what it is, an API to figure out the count of logs in a certain time period, at a certain frequency. Previously it was called `datebins` when it didn't really adhere to the time limits associated with a "date", which is semantically wrong.

<hr>

This PR has:


nikhilsinhaparseable (approved):
looks good

PR #1104, opened 2025-01-16
feat: use `debug_assertions` instead of custom feature

de-sh:
Fixes #XXXX.

### Description

We don't need a custom feature, https://doc.rust-lang.org/reference/conditional-compilation.html#debug_assertions

<hr>

This PR has:


nitisht:
Does this mean the `debug` section is enabled when building without `--release` flag? 

Also can you confirm we don't need a change in the Dockerfile here: https://github.com/parseablehq/parseable/blob/main/Dockerfile.debug

nitisht (commented):
Added a comment

Dockerfile.debug
```diff
@@ -23,7 +23,7 @@ LABEL org.opencontainers.image.licenses="AGPL-3.0"
 
 WORKDIR /parseable
 COPY . .
-RUN cargo build --features debug
```
de-sh, line 26: We have made that change @nitisht

PR #1112, opened 2025-01-17
ci: build for all release targets to ensure PRs are compatible

de-sh:
Fixes #XXXX.

### Description
This changes ensures that we run a CI job to check compatibility of the PR against release targets and ensure we don't fail to compile for a particular target at release.

<hr>

This PR has:

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -35,7 +35,7 @@ jobs:
         run: cargo build --target ${{ matrix.target }}
 
   build-windows:
-    name: Build for ${{matrix.target}}
+    name: Build for windows
     runs-on: windows-latest
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -0,0 +1,76 @@
...
+          toolchain: stable
+          profile: minimal # minimal component installation (ie, no documentation)
+          target: ${{ matrix.target }}
+          override: true
+
+      - uses: actions-rs/cargo@v1
+        with:
+          use-cross: true
+          command: build
+          args: --profile release-lto --target ${{matrix.target}}
```
de-sh, line 38:
```suggestion
          args: --target ${{matrix.target}}
```

nitisht (approved):

PR #1115, opened 2025-01-17
refactor: Correlations

de-sh:
Fixes #XXXX.

### Description

- Use `Deref`, enables changing `self.0` ~> `self`.
- Use `HashMap` to enable quick retreival.
- Encapsulate mutations into a single method.

<hr>

This PR has:


nikhilsinhaparseable:
@de-sh while validating the PR, i found that the id of the correlation is not loaded at the server start but the json has the id.
image
image

can you please check.

nikhilsinhaparseable (approved):
looks good

PR #1117, opened 2025-01-18
fix: oidc connection

nikhilsinhaparseable:
removed requires and group from oidc related args
we are anyways checking in the `openid()` if all three exists
then set OpenidConfig, else None


de-sh:
We are silently not enabling oidc when either of the configs are not present, this is wrong imo, have opened a PR that fixes

PR #1123, opened 2025-01-20
fix: run `fields_mismatch` against `static_schema` data also

de-sh:
Fixes #XXXX.

### Description

Don't cast floats as int on streams with `static_schema_flag` set

<hr>

### Testing Methodology
1. Start parseable.
2. Create new static schema stream with an int field.
3. Push data on the static schema stream with float value.

Without the fix, value is being rounded out, which is wrong as we lose data.

This PR has:


nikhilsinhaparseable (approved):
looks good

PR #1127, opened 2025-01-21
fix: default value when not set

### LINKED ISSUE #1128: Bug: cannot set HotTier
`/about` output:

```json
{
  "version": "1.7.3",
  ...
}
```

Description:

while setting hot tier for a new stream server expects a new field {"used_size"}, where as hot tier was not available for that stream

Image

de-sh:
Fixes #1128.

### Description

API for setting hottier fails as request expects the fields `used_size` and `available_size, which it should not.

## How to Verify
1. Start parseable instance
2. Try to create hottier in UI/API

Before: request fails, now it doesn't

<hr>

This PR has:


Koustavd18 (approved):
LGTM

PR #1129, opened 2025-01-21
test + refactor: staging

de-sh:
Fixes #XXXX.

### Description

While trying to audit the code to perform staging, I made these improvements that reduces unnecessary copies and added tests to critical parts of the flow and hence improve coverage.

<hr>

This PR has:


de-sh:
Only failure is because of alerts check, which will continue to fail until #1101  is merged

parmesant (approved):
LGTM

PR #1130, opened 2025-01-22
fix: fetch first-event-at from storage

nikhilsinhaparseable:
current: query server fetches first-event-at from all the live ingestors 
but, this logic fails when ingestor which ingested the events for a stream is not reachable anymore, query server gets `None`

change: fetch the first-event-at from all the stream jsons from storage,
find the earliest value, and update in server's memory map

### DIFF
--- a/src/catalog/mod.rs
+++ b/src/catalog/mod.rs
@@ -27,6 +27,7 @@ use crate::option::{Mode, CONFIG};
 use crate::stats::{
     event_labels_date, get_current_stats, storage_size_labels_date, update_deleted_stats,
 };
+use crate::storage::object_storage::get_stream_meta_from_storage;
 use crate::{
     catalog::manifest::Manifest,
     event::DEFAULT_TIMESTAMP_KEY,
@@ -279,7 +280,9 @@ async fn create_manifest(
                 }
             };
             first_event_at = Some(lower_bound.with_timezone(&Local).to_rfc3339());
-            if let Err(err) = STREAM_INFO.set_first_event_at(stream_name, first_event_at.clone()) {
+            if let Err(err) =
+                STREAM_INFO.set_first_event_at(stream_name, first_event_at.as_ref().unwrap())
+            {
                 error!(
                     "Failed to update first_event_at in streaminfo for stream {:?} {err:?}",
                     stream_name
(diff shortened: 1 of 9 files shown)


src/storage/object_storage.rs
```diff
@@ -747,3 +783,27 @@ pub fn ingestor_metadata_path(id: Option<&str>) -> RelativePathBuf {
         &format!("ingestor.{}.json", INGESTOR_META.get_ingestor_id()),
     ])
 }
+
+pub async fn get_stream_meta_from_storage(
```
de-sh, line 787: Can we associate this with `ObjectStore`?

src/catalog/mod.rs
```diff
@@ -435,6 +438,58 @@ pub async fn get_first_event(
...
+/// # Examples
+/// ```ignore
+/// ```rust
+/// let result = get_first_event_from_storage("my_stream").await;
+/// match result {
+///     Ok(Some(first_event)) => println!("first-event-at: {}", first_event),
+///     Ok(None) => println!("first-event-at not found"),
+///     Err(e) => eprintln!("Error retrieving first-event-at from storage: {}", e),
+/// }
+/// ```
```
de-sh, line 470: The code comments aren't very helpful, can we restrict to only pointing out working and edge-cases?

src/catalog/mod.rs
```diff
@@ -435,6 +438,58 @@ pub async fn get_first_event(
...
+pub async fn get_first_event_from_storage(
```
de-sh, line 471: Function should be associated to the storage object

src/metadata.rs
```diff
@@ -212,13 +212,50 @@ impl StreamInfo {
...
+    /// # Examples
+    /// ```ignore
+    /// ```rust
+    /// let result = metadata.remove_first_event_at("my_stream");
+    /// match result {
+    ///     Ok(()) => println!("first-event-at removed successfully"),
+    ///     Err(e) => eprintln!("Error removing first-event-at from STREAM_INFO: {}", e),
+    /// }
+    /// ```
+    pub fn reset_first_event_at(&self, stream_name: &str) -> Result<(), MetadataError> {
```
de-sh, line 253: I feel a good comment would explain where we are using this function, e.g. when retention threshold is met and we are removing first event time since that information is outdated.

src/handlers/http/modal/utils/logstream_utils.rs
```diff
@@ -507,3 +508,56 @@ pub async fn create_stream_and_schema_from_storage(stream_name: &str) -> Result<
...
+///```ignore
+/// ```rust
+/// use parseable::handlers::http::modal::utils::logstream_utils::update_first_event_at;
+/// let result = update_first_event_at("my_stream", "2023-01-01T00:00:00Z").await;
+/// match result {
+///     Some(timestamp) => println!("first-event-at: {}", timestamp),
+///     None => eprintln!("Failed to update first-event-at"),
+/// }
+/// ```
+pub async fn update_first_event_at(stream_name: &str, first_event_at: &str) -> Option<String> {
```
de-sh, line 543: Maybe we should pass `DateTime`/`NaiveDateTime` instead of using `&str` to represent this information?

de-sh (approved):

PR #1133, opened 2025-01-24
fix: limit JSON payload size to 10MB

de-sh:
Fixes #XXXX.

### Description
Since #1100 we are using the `Json` extractor, which uses `JsonConfig`, not `PayloadConfig`

### Testing
Send payload of size 5MB to the restricted endpoints, e.g. `/api/v1/ingest`

<hr>

This PR has:


nikhilsinhaparseable (approved):
looks good

PR #1136, opened 2025-01-27
refactor: simply store `StreamType`

de-sh:
Fixes #XXXX.

### Description

No functional changes, just saves on cost of conversion into `String`

<hr>

### How to test
1. Start a parseable instance without this change.
2. Create a stream, push data onto it.
3. Shutdown parseable, update binary with this change.
4. Restart parseable with same config as before, ensure internal streams are considered as such and no conflicts in working.

This PR has:


nikhilsinhaparseable (approved):
looks good

PR #1138, opened 2025-01-28
split arrow conversion and sync

parmesant:
This PR separates out the arrow conversion job from the sync job as we require more visibility into both of them. Arrow to Parquet conversion takes place at the 5th second of every minute and sync/upload happens every 30 seconds. Nothing else has been changed in terms of functionality.

### Description

<hr>

This PR has:


src/storage/object_storage.rs
```diff
@@ -572,13 +573,17 @@ pub trait ObjectStorage: Debug + Send + Sync + 'static {
...
     async fn sync(&self, shutdown_signal: bool) -> Result<(), ObjectStorageError> {
         if !Path::new(&CONFIG.staging_dir()).exists() {
             return Ok(());
         }
 
+        // get all streams
         let streams = STREAM_INFO.list_streams();
 
+        // start the sync loop for a stream
+        // parallelize this
```
nikhilsinhaparseable, line 586: please implement the code for parallelization for each stream in both sync and conversion flow

src/storage/object_storage.rs
```diff
@@ -587,6 +592,9 @@ pub trait ObjectStorage: Debug + Send + Sync + 'static {
                 .get_custom_partition(stream)
                 .map_err(|err| ObjectStorageError::UnhandledError(Box::new(err)))?;
             let dir = StorageDir::new(stream);
+
+            // read arrow files on disk
+            // convert them to parquet
             let schema = convert_disk_files_to_parquet(
```
nikhilsinhaparseable, line 598: why are we still converting arrow to parquet in sync flow

PR #1141, opened 2025-01-29
refactor: only compare names

de-sh:
Fixes #XXXX.

### Description

We are unnecessarily complicating the existence check, this can be simplified using a HashSet as done in the PR.

<hr>

This PR has:


nikhilsinhaparseable (dismissed):
looks good

PR #1161, opened 2025-02-05
fix: kafka build on aarch64

de-sh:
Fixes #XXXX.

### Description

<hr>

This PR has:


hippalus:
@de-sh I found that downgrading Rust from 1.84 to 1.83 in Dockerfile.kafka fixes our issues. While I'm not fully certain about the underlying OpenSSL/libssl compatibility details, using 1.83 appears to be a working solution. Let me know if you'd like me to test anything else.

PR #1182, opened 2025-02-10
fix for converting arrows to parquet

nikhilsinhaparseable:
Fixes #XXXX.

### Description

<hr>

This PR has:

### DIFF
--- a/src/staging/streams.rs
+++ b/src/staging/streams.rs
@@ -165,7 +165,7 @@ impl<'a> Stream<'a> {
             .filter(|file| file.extension().is_some_and(|ext| ext.eq("arrows")))
             .sorted_by_key(|f| f.metadata().unwrap().created().unwrap())
             .collect();
-        
+
         paths
     }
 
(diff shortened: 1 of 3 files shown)


src/staging/streams.rs
```diff
@@ -172,24 +171,36 @@ impl<'a> Stream<'a> {
...
-                    .unwrap()
-                    .to_str()
-                    .unwrap()
-                    .starts_with(&exclude.format("%Y%m%dT%H%M").to_string())
-            });
-        }
+        arrow_files = arrow_files
+            .into_iter()
+            .filter(|path| {
+                let created_at = path.metadata().unwrap().created().unwrap();
```
de-sh, line 187: This will panic if run against a certain few filesystems like btrfs [ref](https://stackoverflow.com/questions/76946130/why-do-i-get-creation-time-is-not-available-on-this-platform-currently-and-wh)

de-sh (commented):
The PR will break support for time-partitioning, file in object store will have file created time and not the timestamp in the log

PR #1211, opened 2025-02-25
fix: capture time created from metadata not filename

de-sh:
Fixes #XXXX.

### Description

Currently we use filename to discover files that aren't to be considered for compaction, with this PR we will consider file creation time instead.

<hr>

This PR has:

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced file exclusion logic by switching to a more robust, system-aware timestamp comparison.
  - Improved precision when determining file creation times to ensure more reliable file processing.


nikhilsinhaparseable (approved):
looks good

PR #1213, opened 2025-02-25
feat: drop use of `MergeReverseRecordReader`

de-sh:
Fixes #XXXX.

### Description

<hr>

This PR has:

## Summary by CodeRabbit

- **New Features**
  - Introduced an enhanced batch iteration method for processing file records more efficiently.

- **Refactor**
  - Streamlined record reading workflows by exclusively using the `MergedRecordReader` and removing legacy reverse processing.
  - Updated instantiation to simplify the creation of `MergedRecordReader` and improved error handling.
  - Simplified iteration over merged records by adjusting method parameters and removing unnecessary complexity.

### DIFF
--- a/src/parseable/staging/reader.rs
+++ b/src/parseable/staging/reader.rs
@@ -29,7 +29,7 @@ use arrow_ipc::reader::StreamReader;
 use arrow_schema::Schema;
 use tracing::error;
 
-use crate::utils::arrow::adapt_batch;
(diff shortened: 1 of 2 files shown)


de-sh:
We won't be proceeding this route because:
> The order of records in response should be time ordered, and not in random order. With this PR we are responding to queries without considering the order.

PR #1220, opened 2025-03-03
fix: multi threaded query execution

nikhilsinhaparseable:
add multi threading execution to query execute method 
allows datafusion to use multiple threads to perform parallel execution of plans 
improves query performance

## Summary by CodeRabbit

- **Refactor**
  - Improved query processing to run in asynchronous mode for enhanced responsiveness.
  - Updated error handling to capture and report issues more effectively.
  - Refined system configuration to optimize concurrent data processing and performance.
  - Introduced new session configuration options for better execution strategies.

- **Chores**
  - Changed default value for row group size to 262144 to optimize data processing.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -264,7 +264,7 @@ pub struct Options {
     #[arg(
         long,
         env = "P_PARQUET_ROW_GROUP_SIZE",
-        default_value = "1048576",
+        default_value = "262144",
         help = "Number of rows in a row group"
     )]
     pub row_group_size: usize,
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -260,7 +260,8 @@ pub struct Options {
         help = "Set a fixed memory limit for query in GiB"
     )]
     pub query_memory_pool_size: Option<usize>,
-
+    // reduced the max row group size from 1048576
+    // smaller row groups help in faster query performance in multi threaded query
     #[arg(
         long,
         env = "P_PARQUET_ROW_GROUP_SIZE",
--- a/src/query/mod.rs
+++ b/src/query/mod.rs
@@ -91,13 +91,11 @@ impl Query {
         let mut config = SessionConfig::default()
             .with_parquet_pruning(true)
             .with_prefer_existing_sort(true)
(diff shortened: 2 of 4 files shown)


src/query/mod.rs
```diff
@@ -145,7 +142,9 @@ impl Query {
             .execute_logical_plan(self.final_logical_plan(&time_partition))
             .await?;
 
-        let fields = df
+        let optimised_df = df.repartition(Partitioning::RoundRobinBatch(16))?;
```
nitisht, line 145: what is 16 here? Can we add a comment

src/query/mod.rs
```diff
@@ -91,27 +91,23 @@ impl Query {
         let mut config = SessionConfig::default()
             .with_parquet_pruning(true)
             .with_prefer_existing_sort(true)
-            .with_round_robin_repartition(true);
+            .with_information_schema(true)
+            .with_batch_size(1000000)
+            .with_coalesce_batches(true);
```
nitisht, line 96: Please add comments on why are we choosing these values and what do they mean for the query

src/query/mod.rs
nikhilsinhaparseable: this is removed, we will utilise all the parallelization available not limit to 16

src/cli.rs
```diff
@@ -264,7 +264,7 @@ pub struct Options {
     #[arg(
         long,
         env = "P_PARQUET_ROW_GROUP_SIZE",
-        default_value = "1048576",
+        default_value = "262144",
```
nitisht, line 267: what are the reasons to change the row group size here - can we add a comment

src/query/mod.rs
```diff
@@ -91,27 +91,20 @@ impl Query {
         let mut config = SessionConfig::default()
             .with_parquet_pruning(true)
             .with_prefer_existing_sort(true)
-            .with_round_robin_repartition(true);
-
+            .with_batch_size(1000000);
```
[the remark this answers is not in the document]
nikhilsinhaparseable: done

nitisht (approved):

PR #1252, opened 2025-03-19
fix: datasets API edgecase where stream is not hottiered

de-sh:
Fixes #XXXX.

### Description

Handles the circumstance where hottier is not configured at the stream level, but hottier is enabled at instance level

<hr>

This PR has:


nikhilsinhaparseable (approved):
looks good

Issue #1290, opened 2025-04-08
Add Kafka Configuration Values to Parseable Helm Template


hippalus:
## Proposed Changes
1. Update the Helm template files to include the following Kafka connector values:
   - `kafkaConnector.bootstrapServers`
   - `kafkaConnector.enabled`
   - `kafkaConnector.consumerTopics`
   - `kafkaConnector.partitionListenerConcurrency`

and some necessary security environments.

2. Ensure that the Kafka connector is only enabled when `kafkaConnector.enabled` is set to `true`.

3. Add appropriate environment variables or configuration mappings in the StatefulSet or Deployment templates to pass these values to the Parseable application.

## Acceptance Criteria
- The Kafka connector values are correctly integrated into the Helm template.
- The Kafka connector is enabled only when `kafkaConnector.enabled` is `true`.
- The Helm chart is tested to ensure the Kafka connector works as expected with the provided values.

PR #1291, opened 2025-04-08
Add mandatory kafka env values to parseable helm template

hippalus:
#1290.

### Description

<hr>

This PR has:

## Summary by CodeRabbit

- **New Features**
  - Enhanced Kafka connector integration with new, configurable environment settings for the deployment.
  - Introduced a new configuration section for customizing Kafka connection settings, consumer topics, and partition concurrency.
  - Added dynamic environment variables for the Kafka connector based on configuration settings.


helm/templates/ingestor-statefulset.yaml
```diff
@@ -139,6 +139,14 @@ spec:
...
           - name: P_MODE
             value: "ingest"
+          {{- if .Values.parseable.kafkaConnector.enabled }}
+          - name: P_KAFKA_BOOTSTRAP_SERVERS
+            value: {{ .Values.parseable.kafkaConnector.bootstrapServers | quote }}
+          - name: P_KAFKA_CONSUMER_TOPICS
+            value: {{ .Values.parseable.kafkaConnector.consumerTopics | quote }}
+          - name: P_KAFKA_PARTITION_LISTENER_CONCURRENCY
+            value: {{ .Values.parseable.kafkaConnector.partitionListenerConcurrency | quote }}
+          {{- end }}
```
AdheipSingh, line 149:
```
{{- if .Values.parseable.kafkaConnector.enabled }}
  {{- range $key, $value := .Values.parseable.kafkaConnector.env }}
  - name: {{ $key }}
    value: {{ tpl $value $ | quote }}
  {{- end }}
{{- end }}
```

If we implement this way, the template will remain cleaner without hardcoding values.

AdheipSingh (changes requested):

helm/templates/ingestor-statefulset.yaml
hippalus: @nitisht @AdheipSingh  Can you check again ?

nitisht (approved):

nitisht:
Thank you @hippalus looks great, merging now

PR #1295, opened 2025-04-11
update user auth for list filters

nikhilsinhaparseable:
current: server checks if user is authorized to a dataset based on query string 
this logic does not work for saved searches because there is no query string 
as searches have key:value pair syntax

update: server fetches the filter_type
if filter_type is sql or filter,
server checks if user is authorized to a dataset based on query string if filter_type is search,
server checks if user is authorized to a dataset based on the dataset_name

## Summary by CodeRabbit

- **New Features**
  - Enhanced filter functionality now categorizes filters into distinct types (e.g., SQL, Search) for clearer processing.
  - Improved asynchronous query handling provides a more responsive experience.

- **Bug Fixes**
  - Addressed inconsistencies in user authorization logic by shifting from query-specific checks to dataset-level permissions.

- **Refactor**
  - Updated user authorization mechanisms across features to enforce streamlined, dataset-level permissions.
  - Reorganized and consolidated internal dependencies for greater system clarity and reliability.

### DIFF
--- a/src/alerts/alerts_utils.rs
+++ b/src/alerts/alerts_utils.rs
@@ -18,7 +18,6 @@
 
 use arrow_array::{Float64Array, Int64Array, RecordBatch};
 use datafusion::{
-    common::tree_node::TreeNode,
     functions_aggregate::{
         count::count,
         expr_fn::avg,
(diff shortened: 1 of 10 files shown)


nitisht:
Merging because tests passed once and we just added comments now.

PR #1306, opened 2025-05-01
fix: delete date stats

parmesant:
Stats were not getting deleted during stream deletion

Fixes #XXXX.

### Description

<hr>

This PR has:

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error messages when deleting local stream data, now providing detailed error information if deletion fails.
- **Refactor**
	- Enhanced metric removal logic to match and delete metrics based on label presence rather than positional prefix.
	- Added safer error handling during metric label removal to prevent error propagation and log warnings instead.


nikhilsinhaparseable (approved):
looks good

PR #1337, opened 2025-06-06
chore: update to helm release v2.3.2

nitisht:
Also update comments in the release action


nitisht:
Skipping tests because this doesn't involve code change.

PR #1350, opened 2025-06-17
chore: update helm chart to v2.3.3

nitisht:
## Summary by CodeRabbit

- **Chores**
	- Updated the Helm chart and application version to 2.3.3.
	- Updated the container image tag to v2.3.3 in the Helm configuration.
	- Added a new Helm chart entry for version 2.3.3 and refreshed creation timestamps for all chart versions.


nitisht:
skipping tests because this is not a code change

PR #1365, opened 2025-07-03
bugfixes for target separation

parmesant:
Fixes include-
- Removal of duplication check
- Proper loading of targets on server init
- Masking of sensitive information (slack URL, alert manager auth credentials)
- Error on modifying name
- Renaming `repeat` to `notificationConfig` and setting least count of interval to 1 minute

### Description

<hr>

This PR has:

## Summary by CodeRabbit

* **New Features**
  * Sensitive information in target data is now masked in all relevant API responses.
  * Added a new resource that is loaded during application initialization.

* **Bug Fixes**
  * Improved error handling for invalid modifications to target configurations, including clearer error messages when attempting to change a target's name.

* **Refactor**
  * Unified and simplified notification configuration fields and interval handling for targets.
  * Updated naming conventions for notification settings to improve clarity.

### DIFF
--- a/src/alerts/target.rs
+++ b/src/alerts/target.rs
@@ -164,10 +164,16 @@ impl Target {
                 })
             }
             TargetType::Other(other_web_hook) => {
+                let endpoint = other_web_hook.endpoint.to_string();
+                let masked_endpoint = if endpoint.len() > 20 {
+                    format!("{}********", &endpoint[..20])
+                } else {
+                    "********".to_string()
+                };
                 json!({
                     "name":self.name,
                     "type":"webhook",
-                    "endpoint":other_web_hook.endpoint,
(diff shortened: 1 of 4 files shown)


src/alerts/target.rs
```diff
@@ -148,16 +140,72 @@ pub struct Target {
...
+                };
+                json!({
+                   "name":self.name,
+                   "type":"slack",
+                   "endpoint":masked_endpoint,
+                   "notificationConfig":self.notification_config,
+                   "id":self.id
+                })
+            }
+            TargetType::Other(other_web_hook) => {
```
nikhilsinhaparseable, line 166: @parmesant are we not masking the webhook endpoint? why is there a different behaviour?
we should mask all endpoints irrespective of the target type.

PR #1387, opened 2025-07-23
chore: upgrade helm chart to latest v2.4.0

nitisht:


nitisht:
skipping tests because not a code change

PR #1412, opened 2025-08-22
Bugfixes for alerts

parmesant:
- respect alert state during modification
- validation for notification snooze time

Fixes #XXXX.

### Description

<hr>

This PR has:

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented scheduling alerts with a UTC timestamp in the past.
  * Updating an alert to Disabled no longer triggers unnecessary task deletion or restart.

* **Improvements**
  * Validation error now explicitly lists valid notification state options (notify, human-time, UTC datetime, indefinite).
  * Modifying disabled alerts skips task-management steps for more predictable, less noisy behavior.


nitisht (changes requested):
Please use proper PR header and commit message

PR #1497, opened 2025-12-23
fix: remove associated filters during stream deletion

spuckhafte:
Fixes #XXXX.

### Description

<hr>

This PR has:

## Summary by CodeRabbit

* **Bug Fixes**
  * Stream deletion now attempts to remove orphaned ("zombie") filters; cleanup failures are logged and no longer cause the overall deletion to fail.

* **New Features**
  * Deletion workflow returns and logs cleanup results with counts of successful and failed filter removals to aid auditing and troubleshooting.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -81,7 +81,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e01ed3140b2f8d422c68afa1ed2e85d996ea619c988ac834d255db32138655cb"
 dependencies = [
  "quote",
- "syn 2.0.108",
+ "syn",
 ]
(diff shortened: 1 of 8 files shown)


spuckhafte (commented):
Let `PARSEABLE::metastore::get_filters` be queryable using logstream name or id.

The current implementation of it fetches all the filters across existing logstreams at once.

parmesant (changes requested):
Let's move filter deletion so that it happens after stream deletion. Also, moving the filter deletion logic in a utility function would be better.

PR #1720, opened 2026-07-11
fix(otel): preserve all exemplars per data point (#1662)

### LINKED ISSUE #1662: Bug: Multiple exemplars per data point overwrite each other in OTel metrics flattening
## Summary

Tracked from PR #1661 (comment: https://github.com/parseablehq/parseable/pull/1661#discussion_r3339400425).

Reported by @parmesant.

## Current Behavior

In `src/otel/metrics.rs`, the `insert_exemplars` helper (introduced in PR #1661 to replace the old `flatten_exemplar`) iterates over the `Vec<Exemplar>` for a data point and writes each exemplar's fields directly into the **same** parent map, using static key names:

- `exemplar_time_unix_nano`
- `exemplar_span_id`
- `exemplar_trace_id`
- `exemplar_value`

Because every exemplar in the slice writes to the same keys, each iteration overwrites the previous one. After the loop only the **last** exemplar is retained in the flattened record. All prior exemplars are silently dropped.

This affects Sum, Histogram, and ExponentialHistogram metric types — all of which carry exemplars as `Vec<Exemplar>`.

The previous implementation (`flatten_exemplar`) returned a `Vec<Map<String, Value>>` (one `Map` per exemplar), which at least preserved each exemplar in a separate structure, though the call sites then merged only the first map.

## Ideal Behavior

Exemplars for a data point should be preserved in full. Possible approaches:

1. **Array column** – Serialize the exemplars as a `Value::Array` of `Value::Object` entries and store them under a single column (e.g., `exemplars`). Each element would contain `exemplar_time_unix_nano`, `exemplar_span_id`, `exemplar_trace_id`, `exemplar_value`, and any filtered attributes. This is the most semantically faithful representation and aligns with how the OTel spec treats exemplars (a list per data point).

2. **Indexed flat columns** – Emit `exemplar_0_time_unix_nano`, `exemplar_1_time_unix_nano`, … etc. This avoids nested types but creates a variable, unbounded number of columns.

3. **Emit one row per exemplar** – Fan out exemplar data so each exemplar becomes its own row joined to its parent data point's fields. This is queryable without nested-type support but multiplies row count.

Option 1 (array column) is recommended as it matches the OTel data model, keeps schema stable, and allows query-layer unnesting when needed.

## References

- PR #1661: Ingestion optimization
- OTel Metrics proto spec: exemplars are `repeated Exemplar` on `NumberDataPoint`, `HistogramDataPoint`, and `ExponentialHistogramDataPoint`.

prabhaks:
## What

Fixes #1662.

`insert_exemplars` in `src/otel/metrics.rs` wrote every exemplar of a data point to the same static columns (`exemplar_time_unix_nano`, `exemplar_span_id`, `exemplar_trace_id`, `exemplar_value`). Since the keys were shared, each exemplar overwrote the previous one and only the last survived. This affected Sum, Histogram and ExponentialHistogram (and Gauge, which shares the same `NumberDataPoint` path).

## Approach

Following the recommended option 1 in the issue, all exemplars for a data point are now serialized into a single nested `exemplars` array column. Each element is an object with `exemplar_time_unix_nano`, `exemplar_span_id`, `exemplar_trace_id`, `exemplar_value` and any filtered attributes. This matches the OTel data model (exemplars are a repeated field per data point) and keeps the schema stable regardless of exemplar count.

### Backward compatibility

Added a `P_OTEL_FLATTEN_EXEMPLARS` flag (default `false`). When set to `true` the previous flat `exemplar_*` columns are emitted (last exemplar wins, same as before) for anyone already querying those columns. Per the discussion on the issue, the default is the new array behavior.

### Bonus fix

Previously exemplar attributes were written straight into the data point map, so they leaked into `compute_series_hash` and fragmented physical-series identity. `exemplars` is now part of `OTEL_METRICS_KNOWN_FIELD_LIST`, so exemplar contents no longer affect the series hash.

## Query validation (PoC)

Before settling on the schema I ran a PoC against the same library versions used here (datafusion 53, arrow/arrow-json 58) to confirm the query layer handles the array-of-objects column. arrow-json infers it as `List<Struct<...>>`, and DataFusion handles reading the column, `array_length`, `unnest`, struct field access, and filtering on nested fields. Details are in the issue thread.

## Testing

- `cargo test`: 311 passed, 0 failed.
- New unit tests cover: all exemplars preserved in the array, legacy flat mode keeps last, empty exemplars insert nothing, end to end through a Sum, exemplar contents do not affect the series hash, and the known-field list update.

## Flag reference

`P_OTEL_FLATTEN_EXEMPLARS` env var / `--otel-flatten-exemplars` flag, default `false` (new nested `exemplars` array). Set to `true` for the legacy flat `exemplar_*` columns.

## Summary by CodeRabbit

* **New Features**
  * Added `--otel-flatten-exemplars` / `P_OTEL_FLATTEN_EXEMPLARS` to control how OTEL metric exemplars are represented when flattening.
  * Default preserves all exemplars in a nested `exemplars` structure; enabling uses legacy flat `exemplar_*` fields (“last wins”).
* **Bug Fixes**
  * Metric series identity/hashing no longer changes based on exemplar contents.
* **Tests**
  * Added regression coverage for nested vs legacy behavior, empty exemplars, end-to-end flattening, and series-hash invariance.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -292,6 +292,14 @@ pub struct Options {
     )]
     pub metrics_endpoint_auth: bool,
 
+    #[arg(
+        long,
+        env = "P_OTEL_FLATTEN_EXEMPLARS",
+        default_value = "false",
+        help = "Use the legacy flat per-exemplar columns (exemplar_time_unix_nano, exemplar_span_id, exemplar_trace_id, exemplar_value) for OTEL metrics. These columns keep only the last exemplar of each data point. When false (default), all exemplars are stored under a single nested `exemplars` array column."
+    )]
+    pub otel_flatten_exemplars: bool,
+
     // TLS/Security
     #[arg(
         long,
--- a/src/otel/metrics.rs
+++ b/src/otel/metrics.rs
@@ -132,15 +132,23 @@ fn compute_series_hash(dp: &Map<String, Value>) -> u64 {
     hasher.finish()
 }
 
-/// Convert a single exemplar's value into a JSON number, mirroring the
-/// int/double handling used for data point values.
+/// Convert a single exemplar's value into a JSON number.
+///
+/// Both int and double exemplar values are emitted as `f64` so the flattened
+/// `exemplars_exemplar_value` column has a stable `List<Float64>` type. If int
+/// and double values were emitted as-is, the column would infer as
+/// `List<Int64>` in some batches and `List<Float64>` in others; Parseable's
+/// schema-conflict handling would then route the mismatched batch to a separate
+/// `exemplars_exemplar_value_list` column, leaving the original empty
+/// (reported in the PR #1720 review).
 fn exemplar_value_to_json(value: &ExemplarValue) -> Value {
-    match value {
-        ExemplarValue::AsDouble(double_val) => serde_json::Number::from_f64(*double_val)
-            .map(Value::Number)
-            .unwrap_or(Value::Null),
-        ExemplarValue::AsInt(int_val) => Value::Number(serde_json::Number::from(*int_val)),
-    }
+    let as_f64 = match value {
+        ExemplarValue::AsDouble(double_val) => *double_val,
+        ExemplarValue::AsInt(int_val) => *int_val as f64,
+    };
+    serde_json::Number::from_f64(as_f64)
+        .map(Value::Number)
+        .unwrap_or(Value::Null)
 }
 
 /// Build one JSON object for a single exemplar. Used to populate the nested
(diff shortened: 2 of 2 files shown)


prabhaks:
Note on the failing `coverage` check: it fails at the clippy step because of pre-existing lints on `main` from the newer Rust/clippy 1.97, not because of this PR (none of the flagged files are touched here). I opened #1721 to fix those lints and unblock CI. Once #1721 is merged I will rebase this branch on top so its checks go green.

nikhilsinhaparseable:
@prabhaks thank you for raising the lint PR, i have verified and merged it, you can rebase this PR now

prabhaks:
> @prabhaks thank you for raising the lint PR, i have verified and merged it, you can rebase this PR now

Done, thank you for reviewing and merging my other PR!

prabhaks:
@nikhilsinhaparseable @parmesant please take a look at this PR whenever you get a chance!

prabhaks:
@parmesant @nikhilsinhaparseable Any thoughts on this fix?

parmesant:
Hey @prabhaks 
Got caught up with another issue today. I'll pick this one up tomorrow

parmesant:
@prabhaks the PR looks good but there is a doubt that we have-
- While ingesting data from `astronomy shop`, we could see that a new column called `exemplars_exemplar_value_list` got created which contained the array of values and another column `exemplars_exemplar_value` which is empty. Is that intended?

prabhaks:
> @prabhaks the PR looks good but there is a doubt that we have-
> 
> * While ingesting data from `astronomy shop`, we could see that a new column called `exemplars_exemplar_value_list` got created which contained the array of values and another column `exemplars_exemplar_value` which is empty. Is that intended?

Hi @parmesant - I am on vacation till July 22nd. Will get back to you after that!

parmesant:
Have a great vacation!

prabhaks:
Good catch, and no, that empty column is not intended.

Root cause: exemplar values come through as an integer for `AsInt` exemplars and a float for `AsDouble`. Once the `exemplars` array is flattened on ingest, each field becomes a list column, so `exemplars_exemplar_value` infers as `List<Int64>` for some batches and `List<Float64>` for others. When the inferred type differs from what the stream already stored, the conflict handling routes that batch into a typed sibling column, which is where `exemplars_exemplar_value_list` comes from, and the original `exemplars_exemplar_value` is left empty. The astronomy shop mixes int and double exemplars, so it runs into this.

I reproduced it with a small test: an int-typed exemplar batch followed by a double-typed one yields exactly `{"exemplars_exemplar_value": "exemplars_exemplar_value_list"}` as a conflict, while double followed by double yields no conflict.

Fix (just pushed): normalize both int and double exemplar values to `f64` in the flattener, so `exemplars_exemplar_value` is always `List<Float64>`. That gives a single clean column with no empty sibling. The other exemplar fields (span_id, trace_id, time) are strings and were already stable.

One heads up for re-testing: the earlier run already created the empty `exemplars_exemplar_value` and the `exemplars_exemplar_value_list` columns in that stream's schema, so you will want to test against a fresh stream (or recreate it) to avoid looking at the columns from the old schema.

For context on the overall shape: the exemplars array flattens into parallel per-field list columns (`exemplars_exemplar_time_unix_nano`, `exemplars_exemplar_span_id`, `exemplars_exemplar_trace_id`, `exemplars_exemplar_value`, plus any filtered attributes), correlated by position within each data point. Happy to adjust the layout if you would prefer something different.

prabhaks:
@parmesant Can you please test against a fresh stream!

parmesant:
Sure
Will do

prabhaks:
> Sure Will do

thank you @parmesant , please let me know your findings!

parmesant:
@prabhaks the PR looks good. Just made one small change to exclude the flattened `exemplars_*` columns from series hash calculations

prabhaks:
> @prabhaks the PR looks good. Just made one small change to exclude the flattened `exemplars_*` columns from series hash calculations

thanks @parmesant for verifying and making the change! If everything looks fine, please approve the PR, so that I can merge it!

parmesant (approved):

PR #1724, opened 2026-07-12
fix(otel): sync OTEL_METRICS_KNOWN_FIELD_LIST with emitted keys (#1723)

### LINKED ISSUE #1723: Bug: OTEL_METRICS_KNOWN_FIELD_LIST out of sync with emitted keys (metric series fragmentation and collision)
## Summary

`OTEL_METRICS_KNOWN_FIELD_LIST` in `src/otel/metrics.rs` is out of sync with the keys the metric flatteners actually emit. `compute_series_hash` treats every top-level key that is NOT in this list as a series label, so the mismatch corrupts physical-series identity for metrics in two opposite ways: some real per-sample fields are treated as labels (fragmenting a series into many), and some names that are never emitted are listed (which can merge distinct series if a user attribute happens to use one of those names).

## Background

`compute_series_hash` (`src/otel/metrics.rs`) hashes `metric_name` plus every top-level key/value on the flattened data point that is not present in `OTEL_METRICS_KNOWN_FIELD_LIST`. The result is stored per sample as `__series_hash` and used by the query layer to group samples into series. So the list must exactly match the set of non-label (sample/metric-level) keys the flatteners emit.

## Problem 1: emitted keys missing from the list (series fragmentation)

These are emitted as top-level keys but are absent from the list, so their per-sample values get hashed into the series identity. Samples of the same physical series then get different `__series_hash` values and are split into many spurious series.

- `min` — emitted by `flatten_histogram` (around line 351, via `insert_number_if_some`)
- `max` — emitted by `flatten_histogram` (around line 352)
- `data_point_quantile_values` — emitted by `flatten_summary` (around line 471) as a nested array of `{quantile, value}` objects

Impact: histogram series fragment whenever `min`/`max` change across scrapes (routine, especially for DELTA temporality), and summary series fragment as quantile values change every sample.

## Problem 2: listed names that are never emitted (dead entries, with collision risk)

These appear in the list but are never emitted as top-level keys, so they exclude any real attribute of the same name from the series hash:

- `value` (list) — only ever emitted nested inside the `data_point_quantile_values` objects (around line 484), never as a top-level key. `value` is a very common attribute name, so a user label named `value` would be silently dropped from series identity and distinct series would collide.
- `quantile` (list) — same situation (nested only, around line 479); collision risk for an attribute named `quantile`.
- `data_point_quantile_values_quantile` and `data_point_quantile_values_value` (list) — never emitted anywhere. Harmless dead entries (names are specific enough to be unlikely user labels), but they should be removed for correctness.

## Proposed fix

- Add `min`, `max`, and `data_point_quantile_values` to `OTEL_METRICS_KNOWN_FIELD_LIST`.
- Remove `value`, `quantile`, `data_point_quantile_values_quantile`, and `data_point_quantile_values_value`.
- Update the list length constant and the `OTEL_METRICS_KNOWN_FIELD_LIST.len()` test assertion accordingly.
- Add tests asserting that differing `min`/`max` (histogram) and `data_point_quantile_values` (summary) do not change `compute_series_hash`, and that a data point attribute named `value` or `quantile` does change it.

## Related

Same class of issue as exemplar attributes leaking into the series hash, discussed in #1662 (see also #1720). Found while auditing the flattening code after that fix.

prabhaks:
## What

Fixes #1723.

`compute_series_hash` in `src/otel/metrics.rs` treats every top-level data-point key that is not in `OTEL_METRICS_KNOWN_FIELD_LIST` as a series label. The list had drifted from the keys the flatteners actually emit, in both directions, corrupting physical-series identity for metrics.

### Emitted but missing from the list (series fragmentation)

These are emitted as top-level keys but were absent from the list, so their per-sample values were hashed into `__series_hash`, splitting one physical series into many:

- `min`, `max` — histogram per-sample statistics (`flatten_histogram`)
- `data_point_quantile_values` — summary quantile array (`flatten_summary`)

### Listed but never emitted top-level (dead entries, with collision risk)

- `value`, `quantile` — only ever emitted nested inside the `data_point_quantile_values` objects, never as top-level keys. Because they were listed, a user attribute literally named `value` or `quantile` was silently excluded from series identity, merging distinct series. `value` is a common label name, so this was the higher-risk one.
- `data_point_quantile_values_quantile`, `data_point_quantile_values_value` — never emitted anywhere. Harmless dead entries, removed for correctness.

## Change

- Add `data_point_quantile_values` to the list.
- Remove `value`, `quantile`, `data_point_quantile_values_quantile`, `data_point_quantile_values_value`.
- Rename the emitted histogram statistics from bare `min`/`max` to `data_point_min`/`data_point_max`, matching the naming convention every other data-point field already uses (`data_point_value`, `data_point_count`, `data_point_sum`, etc.), and add both to the list under their new names. This was raised in review: bare `min`/`max` would have shadowed a real user attribute of the same name and wrongly excluded it from the series hash, the same collision risk as `value`/`quantile` above.
- List length 38 -> 37 (relative to `main` post-#1720, which already carries `exemplars`).
- Tests: differing `data_point_min`/`data_point_max` (histogram) and `data_point_quantile_values` (summary) no longer change `compute_series_hash`; a data-point attribute named `value`, `quantile`, `min`, or `max` still does; plus a guard asserting the list contents and length.

## Verification

- `cargo test --lib otel::metrics`: 18 passed, 0 failed.
- `cargo clippy --all-targets -- -D warnings`: clean on this file.
- Cross-checked the full set of top-level keys emitted by every flattener (number/gauge/sum/histogram/exp-histogram/summary, exemplars, flags, aggregation temporality, metric-level and resource/scope envelope fields) against the updated list: they now correspond 1:1, with no missing keys and no dead entries.
- Rebased cleanly onto `main` after #1720 merged.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected metric field names for sample-level quantiles, minimums, and maximums.
  - Updated histogram data to consistently use prefixed minimum and maximum fields.
  - Removed obsolete flattened metric fields.

- **Tests**
  - Expanded coverage for metric series hashing and known-field membership.
  - Added validation for the updated set of 39 known metric fields.

### DIFF
--- a/src/otel/metrics.rs
+++ b/src/otel/metrics.rs
@@ -37,7 +37,7 @@ use super::otel_utils::{
     convert_epoch_nano_to_timestamp, insert_attributes, insert_number_if_some,
 };
 
-pub const OTEL_METRICS_KNOWN_FIELD_LIST: [&str; 38] = [
+pub const OTEL_METRICS_KNOWN_FIELD_LIST: [&str; 37] = [
     "metric_name",
     "metric_description",
     "metric_unit",
@@ -57,8 +57,15 @@ pub const OTEL_METRICS_KNOWN_FIELD_LIST: [&str; 38] = [
     "data_point_sum",
     "data_point_bucket_counts",
     "data_point_explicit_bounds",
-    "data_point_quantile_values_quantile",
-    "data_point_quantile_values_value",
+    // Nested array of summary quantile objects, each carrying `quantile`
+    // and `value`. Stored under this single key and kept out of the
+    // series hash so per-sample quantile values never fragment series
+    // identity. (The bare `quantile`/`value` keys only ever appear nested
+    // inside this array, never as top-level data-point fields.)
+    "data_point_quantile_values",
+    // Histogram per-sample min/max statistics.
+    "min",
+    "max",
     "data_point_scale",
     "data_point_zero_count",
     "data_point_flags",
@@ -67,8 +74,6 @@ pub const OTEL_METRICS_KNOWN_FIELD_LIST: [&str; 38] = [
     "positive_bucket_count",
     "negative_offset",
     "negative_bucket_count",
-    "quantile",
-    "value",
     "is_monotonic",
     "aggregation_temporality",
     "aggregation_temporality_description",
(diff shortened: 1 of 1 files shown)


prabhaks:
@parmesant @nikhilsinhaparseable Please review this issue and fix!

nitisht:
@prabhaks can you address the conflict pls

prabhaks:
Good catch, thanks. Bare min/max would have shadowed a real user attribute with the same name and wrongly excluded it from the series hash, the same class of bug this PR is meant to fix for value/quantile.

Renamed the emitted keys to data_point_min/data_point_max, matching the existing naming convention for every other data point field (data_point_value, data_point_count, data_point_sum, etc), and added a regression test that a user attribute literally named min or max still affects the series hash. Pushed in e417c87.

prabhaks:
@parmesant  whenever you get time - please look into this!

nikhilsinhaparseable (approved):

Issue #1763, opened 2026-08-21
Deletion of large datasets takes a while


parmesant:
If a user has a large dataset worth TBs stored in any `ObjectStore` like `S3`, `GCS`, etc. its deletion takes a while and the user is still able to see the dataset in the UI.

Presently, the deletion API tries to delete the dataset files from object store and then respond back to the UI with a 200 OK

Proposed solution: make a `.tombstone` file while marks a dataset for deletion and prevents any new actions from being performed on that dataset.

For example
- User sends a deletion request for dataset A
- Query node creates a `.tombstone` file in the root of that dataset and informs other ingest nodes of the same and sends back a 200 OK to the user. It also kicks off a background task in charge of deletion.
- All the nodes delete references to the dataset from their memory while the query node keeps deleting in the background
- During live ingestion, the ingestor should be able to deny incoming requests saying the dataset is being deleted 
- Once the deletion is complete (the background job returns 200 OK), the query node can delete the `.tombstone` file and the folder
- If the query node restarts in the middle of deletion, while loading the datasets it will read the `.tombstone` file and start deleting in background without loading the dataset

prabhaks:
Hi @parmesant @nitisht  I would like to a stab on this issue! Can you please assign it to me?

PR #1768, opened 2026-08-26
Add tombstone/deleting safety net for stream deletion (#1763)

prabhaks:
## Summary

First of two PRs toward fixing #1763 (deletion of large datasets takes a while). This PR only lays the safety-net groundwork so a later PR can make stream deletion asynchronous without races. It does not change the behavior of the current delete handlers, since nothing here actually starts writing a tombstone or setting the `deleting` flag yet.

- `TOMBSTONE_ROOT_DIRECTORY` (`.tombstones`) plus `tombstone_path`/`is_tombstoned` helpers in the storage layer. The tombstone lives outside the `{tenant}/{stream}` prefix that a bulk stream delete walks, so a mid-deletion crash can't lose the marker.
- `LogStreamMetadata.deleting` (transient, in-memory only) and `Stream::mark_deleting`/`is_deleting`.
- A tombstone check as the first step of `create_stream_and_schema_from_storage`, so a concurrent lazy reload can't resurrect a stream that's mid-deletion.
- An unconditional `is_deleting()` check in `create_streams_for_distributed` and in the `logstream` info/schema/stats endpoints, so an already-resident stream flagged `deleting` is rejected instead of served.
- `PostError::StreamBeingDeleted` -> 409 Conflict, checked in `validate_stream_for_ingestion`.
- `.tombstones` added to both `list_streams()` implementations' directory-exclusion filters.

The actual behavior change (tombstone-then-background-delete rewrite of the delete handlers, restart recovery, ingestor self-heal) is scoped as a follow-up PR once this lands, since it needs live-cluster validation that this additive piece doesn't.

## Test plan

- [x] `cargo build --lib`
- [x] `cargo test --lib` (442 passed, 1 pre-existing ignore, 0 failed)
- [x] `cargo fmt --check`
- [x] `cargo clippy --lib --all-targets` (no new warnings; all 10 pre-existing warnings are confined to `src/event/format/known_schema.rs`)
- [x] New unit tests: `is_tombstoned`/`tombstone_path` against a real `LocalFS` backend, `Stream::mark_deleting`/`is_deleting`

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented ingestion, querying, and stream information requests from proceeding while a stream is being deleted.
  * Improved handling of deletion-in-progress responses with clear conflict and not-found errors.
  * Prevented deleted streams and deletion markers from reappearing during storage discovery or lazy loading.
  * Improved query time-range handling for end-time expressions.
* **Tests**
  * Added coverage for stream deletion-state transitions and storage deletion markers.

### DIFF
--- a/src/handlers/http/ingest.rs
+++ b/src/handlers/http/ingest.rs
@@ -553,6 +553,8 @@ pub enum PostError {
     MissingQueryParameter,
     #[error(transparent)]
     MetastoreError(#[from] MetastoreError),
+    #[error("Stream {0} is being deleted, please retry after some time")]
+    StreamBeingDeleted(String),
 }
 
 impl actix_web::ResponseError for PostError {
@@ -585,6 +587,8 @@ impl actix_web::ResponseError for PostError {
             | JsonFlattenError(_) => StatusCode::INTERNAL_SERVER_ERROR,
 
             StreamNotFound(_) => StatusCode::NOT_FOUND,
+
+            StreamBeingDeleted(_) => StatusCode::CONFLICT,
 
             MetastoreError(e) => e.status_code(),
         }
--- a/src/handlers/http/logstream.rs
+++ b/src/handlers/http/logstream.rs
@@ -186,6 +186,9 @@ pub async fn get_schema(
     }
 
     let stream = PARSEABLE.get_stream(&stream_name, &tenant_id)?;
+    if stream.is_deleting() {
+        return Err(StreamNotFound(stream_name.clone()).into());
+    }
     match update_schema_when_distributed(&vec![stream_name.clone()], &tenant_id).await {
         Ok(_) => {
             let schema = stream.get_schema();
@@ -313,6 +316,12 @@ pub async fn get_stats(
     {
         return Err(StreamNotFound(stream_name.clone()).into());
     }
+    if PARSEABLE
+        .get_stream(&stream_name, &tenant_id)
+        .is_ok_and(|stream| stream.is_deleting())
+    {
+        return Err(StreamNotFound(stream_name.clone()).into());
+    }
 
     let query_string = req.query_string();
     if !query_string.is_empty() {
@@ -375,6 +384,12 @@ pub async fn get_stream_info(
     if !PARSEABLE
         .check_or_load_stream(&stream_name, &tenant_id)
         .await
+    {
+        return Err(StreamNotFound(stream_name.clone()).into());
+    }
+    if PARSEABLE
+        .get_stream(&stream_name, &tenant_id)
+        .is_ok_and(|stream| stream.is_deleting())
     {
         return Err(StreamNotFound(stream_name.clone()).into());
     }
--- a/src/handlers/http/modal/utils/ingest_utils.rs
+++ b/src/handlers/http/modal/utils/ingest_utils.rs
@@ -509,6 +509,10 @@ pub fn validate_stream_for_ingestion(
 ) -> Result<(), PostError> {
     let stream = PARSEABLE.get_stream(stream_name, tenant_id)?;
 
+    if stream.is_deleting() {
+        return Err(PostError::StreamBeingDeleted(stream_name.to_string()));
+    }
+
     // Validate that the stream's log source is compatible
     stream
         .get_log_source()
--- a/src/handlers/http/query.rs
+++ b/src/handlers/http/query.rs
@@ -559,6 +559,22 @@ pub async fn create_streams_for_distributed(
     streams: Vec<String>,
     tenant_id: &Option<String>,
 ) -> Result<(), QueryError> {
+    // A stream that's already resident in memory but flagged `deleting`
+    // must reject the query outright. Checked unconditionally, ahead of
+    // the mode gate below, since this function backs every query-side
+    // call site (ad-hoc queries, alerts, saved query context, traces),
+    // not just the querier's own reload path.
+    for stream_name in &streams {
+        if PARSEABLE.streams.contains(stream_name, tenant_id)
+            && let Ok(stream) = PARSEABLE.get_stream(stream_name, tenant_id)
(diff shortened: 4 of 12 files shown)


prabhaks:
Thanks for the review — went through all 5 findings against the actual code. Pushed a fix for 4 of them (commit 094389a on this branch); one I'm leaving as a documented, accepted limitation.

**Fixed:**
- `src/parseable/streams.rs` — `set_metadata` was replacing the whole `LogStreamMetadata`, so a reload racing a delete could silently clear `deleting` back to `false` despite it being documented as monotonic. Now preserves it (`updated_metadata.deleting |= metadata.deleting`), matching your suggested diff.
- `src/storage/object_storage.rs` — `list_tombstoned_streams` returned every directory `list_dirs_relative` found under the tombstone root without confirming the actual `.tombstone` marker existed inside each one. Now re-verifies each candidate with `is_tombstoned` before reporting it, with a regression test for a marker-less directory.
- `src/storage/localfs.rs` — added `TOMBSTONE_ROOT_DIRECTORY` to `list_old_streams`'s ignore list too, for consistency with `list_streams`. Confirmed `list_old_streams` has no callers anywhere in this codebase currently, so this was latent rather than reachable, but no reason to leave it inconsistent.
- `src/parseable/mod.rs` (`check_or_load_stream`) — rather than adding an `is_tombstoned()` I/O check to the resident-stream fast path (this function is on the hot path for essentially every query/schema/stats/hot-tier request, so an object-store round trip there for every access is a real cost for closing a millisecond-scale window), I closed the practical version of this race at the source: the follow-up PR (#1770) now flips `mark_deleting()` before the tombstone write, with no `.await` between them, so there's no window on the initiating node where the tombstone is durable but the local flag isn't set yet. Cross-node propagation is still via the existing push/self-heal (not synchronous) — an accepted, already-documented limitation of this design, not something either PR attempts to fully close.

**Left as-is, with reasoning:** `Stream::push` not checking `is_deleting()` at the write boundary. `validate_stream_for_ingestion` already checks `is_deleting()` before processing starts, so this is a TOCTOU backstop for a delete racing in during that narrow window, not the primary guard. Fully closing it means coordinating `push()` and `mark_deleting()` under a shared lock on the hot ingestion write path — a real architectural change, and the residual risk (a stray write landing right as a delete starts) is the same class as the already-documented, out-of-scope "orphan file" race in the linked issue, not new. Tracking it there rather than taking on that lock coordination in this PR.

prabhaks:
@parmesant cam you review this scaffolding, the actual change is in #1770 which I am still testing, and probably need your help too!

parmesant:
> @parmesant cam you review this scaffolding, the actual change is in #1770 which I am still testing, and probably need your help too!

Sure! Give me till tomorrow to get back on this.

prabhaks:
> > @parmesant cam you review this scaffolding, the actual change is in #1770 which I am still testing, and probably need your help too!
> 
> Sure! Give me till tomorrow to get back on this.

thanks @parmesant !

prabhaks:
**How the tombstone-based deletion works**

The core problem: today, `DELETE /logstream/{stream}` blocks the HTTP response on a full recursive object-store delete. For a TB-scale stream that's potentially millions of keys, so the client waits minutes for something that should be instant.

**The fix, in three parts:**

1. **Durable marker, placed outside the stream's own prefix.** When a delete request comes in, we write a tiny marker object to `.tombstones/{tenant}/{stream_name}/marker` — deliberately *not* under the stream's own `{tenant}/{stream_name}/...` prefix. That matters because the actual bulk delete is a single recursive LIST-then-DELETE over that exact prefix. If the tombstone lived inside it, the bulk delete could sweep it up mid-job, and then a crash right after would leave no record that a deletion was ever in progress — breaking the "resume on restart" guarantee. Placing it outside makes it structurally immune to that, regardless of backend or listing order.

2. **In-memory `deleting` flag, set before the tombstone write.** The moment a delete request lands, we flip a `deleting` bool on the resident `Stream` object in memory (not persisted — it's re-derived from the tombstone on reload). This flag is checked at every place a stream could be touched: query execution, ingestion, schema/stats lookups, and stream reload from storage. It's set *before* the tombstone `put_object` call completes, with no `await` in between, specifically to close a race where a concurrent request on the same node could otherwise slip through in the gap between "tombstone durable" and "flag set."

3. **Respond immediately, delete in the background.** Once the tombstone is written and the flag is set, we respond `202 Accepted` right away. The actual bulk delete runs in a spawned background task. When it finishes, it clears the tombstone and removes the stream from memory. If the node crashes mid-delete, the tombstone survives (per point 1), and on restart we scan `.tombstones/` and resume the deletion for anything still marked — so there's no orphaned half-deleted stream state after a crash.

**Cross-node correctness (this is a distributed system, not a single process):**
- The node that receives the DELETE (query node, or standalone) is the only one that ever runs the actual physical bulk delete.
- It notifies all live ingestors synchronously (before responding) to flag the stream `deleting` locally and stop new writes to it.
- As a fallback for an ingestor that was down or partitioned when that notification went out, the periodic sync job also checks for the tombstone and self-heals — so the worst case is bounded to one sync interval, not indefinite staleness.
- Only the originating node's background job ever runs the actual delete-stream call — ingestors just flag and wait, they never independently re-trigger the bulk delete.

**Split into two PRs:**
- **#1768** is pure scaffolding — tombstone helpers, the `deleting` flag, and guards everywhere that checks it — with zero change to today's delete handlers' actual behavior.
- **#1770** is what wires it up: writes the tombstone, sets the flag, fans out to ingestors, spawns the background delete, and changes the response to 202.

Happy to expand on any specific part (crash-recovery ordering, ingestor self-heal timing, why 202 vs 200, etc.) if useful.

parmesant (approved):

prabhaks:
@parmesant can you please re-review this?

parmesant (approved):

